PR #15, opened 2016-05-18
Make sure video view panel stays on top

### LINKED ISSUE #1: Ensure vid-view stays on top when switching tabs
Make sure the video view is not dismissed when interacting with page.

The current implementation uses `sdk/panel`. The default actions for this component mean it is dismissed with `<esc>` or body:click. We need to do some research and see what type of component we should use instead.

meandavejustice:
- fixes #1
- bump version to `0.0.2`
- update packaged xpi


6a68:
looking now

6a68:
this doesn't seem to work for me, opening a counter PR that does seem to work :-)

PR #18, opened 2016-05-18
fix typo: height

pdehaan:
Unrelated: Not sure how tied we are to FF 38.x support, but if we want to jump on the ES6 template string bandwagon we could possibly micro optimize some of the string concat stuff.


meandavejustice:
Not sure what the story around support is atm. I'm cool with using template strings and back porting if we need to down the road.

PR #24, opened 2016-05-18
Disable panel autohide

### LINKED ISSUE #1: Ensure vid-view stays on top when switching tabs

6a68:
Fixes #1.

@meandavejustice mind taking a look? this solves the problem of the panel autohiding itself, but introduces the problem that you can't get rid of the panel. :-)

The fix is to call panel.hide(), I'm just not yet sure how we can add a button to the panel / style the panel.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,9 @@ var panel = require("sdk/panel").Panel({
   }
 });
 
+var { getActiveView } = require("sdk/view/core");
+getActiveView(panel).setAttribute("noautohide", true);
+
 var cm = require("sdk/context-menu");
 
 cm.Item({


6a68:
Hmm, I think I'm unlocking some XUL secrets. Don't merge this just yet.

6a68:
Ah, never mind, I guess I"ll have to fix closing and moving the panel in a separate PR. `backdrag=true` seems to be the attribute used by XUL to make an element draggable, but for some reason it's not working in this case. Not sure why.

meandavejustice:
I address this in #15. I also tested the backdrag, and I wasn't able to get it work either.

meandavejustice:
Thank you!

PR #30, opened 2016-05-25
Adds keys to generate update.rdf

kaganjd:


meandavejustice:
@kaganjd I think we should hold off on this until we know where we will be hosting the min-vid addon. More than likely it will be on s3

6a68:
Closing this for now, we'll revisit when we figure out the S3 stuff

Issue #13, opened 2016-05-13
Automate uploads to s3


meandavejustice:
What is our process for deploying/hosting the packaged and signed addon?
- amo?
- gh-releases?
- gh-pages?
- people.mozilla.org?

6a68:
Maybe just use the same S3 approach that one of the existing add-ons is using? Might be really nice to create some kind of command-line utility that test pilot add-ons (or other non-AMO-hosted add-ons) can reuse.

wresuolc:
Let's use S3 w/ a script to sign from AMO and upload automatically

meandavejustice:
example scripts => https://github.com/mozilla/testpilot/tree/master/addon/bin

pdehaan:
FWIW, here's Activity Stream's fabfile for XPI signing and S3 deploying. Uses Node+Python, which may not be the most ideal solution here: https://github.com/mozilla/activity-stream/blob/master/fabfile.py

meandavejustice:
I have received my aws keys, so this is back in progress

wresuolc:
This is not important for the launch.  We can always do this manually.

meandavejustice:
Signing is complete, right now we are handing off manually to wil.

PR #34, opened 2016-05-26
add dist to gitignore

johngruen:


meandavejustice:
Going to close this until we figure out #13. It's much easier for me to point people to the xpi bundled here than email copies around.

PR #36, opened 2016-05-27
Fixes #27 Doesn't work on youtube.com/watch pages

### LINKED ISSUE #27: Doesn't work on youtube.com/watch pages
Youtube shows a custom context menu if you right-click over the youtube.com/watch player (see screenshot below).

Maybe we could have a toolbar button light up if the window.location is youtube.com/watch?..., since it'd be easy to grab the window.location and send that to the picture-in-picture player.

screen shot 2016-05-20 at 12 04 56 pm

kaganjd:
Within cm.SelectorContext, this uses "class*=" instead of "href*=" to trigger the context menu on YouTube.com main pages and channel pages.

For some reason, using "href*=/watch?v" wasn't working.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,6 +49,20 @@ cm.Item({
   }
 });
 
+cm.Item({
+  label: "Send to mini player",
+  context: [
+  cm.URLContext(['*.youtube.com']),
+  cm.SelectorContext('[class*="yt-uix-sessionlink"]'),
+  ],
+  contentScript: "self.on('click', function (node, data) {" +
+                 "  self.postMessage(node.href);" +
+                 "});",
+  onMessage: function(url) {
+    updatePanel(constructYoutubeEmbedUrl(url));
+  }  
+});
+
 function updatePanel(url) {
   panel.port.emit('set-video', url);
   panel.show();


index.js
```diff
@@ -44,7 +44,7 @@ var cm = require("sdk/context-menu");
 
 cm.Item({
   label: "Send to mini player",
-  context: cm.SelectorContext('[href*="youtube.com"], [href*="youtu.be"]'),
+  context: cm.SelectorContext('[class*="yt-uix-sessionlink"], [href*="youtube.com"], [href*="youtu.be"]'),
```
meandavejustice: I think that checks for classname and attributes specific to the youtube domain should be in a separate call to `cm.Item`

This would probably look something like:

``` js
cm.Item({
  label: "Send to mini player",
  context: [
    cm.URLContext(['*.youtube.com']),
    cm.SelectorContext('href*="watch"')
  ],
```

kaganjd: That makes sense. Thanks! 

New commit below. Should I squash them all into one?

meandavejustice: squashing would be great, thank you!

PR #37, opened 2016-05-27
Adds a Development section and steps for getting started

kaganjd:
Also expands on description of the add-on.


README.md
```diff
@@ -1,18 +1,42 @@
...
+
+1.  Clone the repo:
+
+`https://github.com/meandavejustice/min-vid.git`
+
+2.  Install packages:
+
+`npm install` 
+
+3.  Install jpm, the utility for Mozilla Add-ons:
```
meandavejustice: You can leave this step out since `jpm` is listed as a `devDependency` in package.json

README.md
```diff
@@ -1,18 +1,42 @@
...
+4. Before submitting a Pull Request, package the Add-on:
+
+`npm run package`
+
+## Helpful
+Use jpm with the [Extension Auto-Installer](https://addons.mozilla.org/en-US/firefox/addon/autoinstaller/) to automatically update your add-on when you make changes.
+
+After you install Extension Auto-Installer, run:
+
+`jpm watchpost --post-url http://localhost:8888/`
```
meandavejustice: We should instruct users to use the `npm run watch` command here. That way they don't have to install jpm globally. We should definitely keep the above note about installing Extension Auto-Installer though

README.md
```diff
@@ -1,18 +1,42 @@
-# min-vid
-a firefox addon which lets you minimize web videos
+# Min-Vid
+Min-Vid is a Firefox add-on which lets you minimize a web video and keep it within the browser window. The video panel stays visible even when you switch tabs, so you can keep watching while you browse.
```
meandavejustice: Thanks for cleaning this up!

README.md
```diff
@@ -1,18 +1,38 @@
...
+
+`https://github.com/meandavejustice/min-vid.git`
+
+2.  Install packages:
+
+`npm install` 
+
+3. Before submitting a Pull Request, package the Add-on:
+
+`npm run package`
```
pdehaan: We may need to tweak some indentation here. Looking at the rich diff on GitHub, it looks like the numbers reset to 1. for each step. Markdown is odd and not always intuitive.

kaganjd: Ah, thanks for catching that. I deleted this branch because it turned into a giant mess. Will make the indentation change on the new PR and squash the commits.

Issue #8, opened 2016-05-13
Figure out metrics story


meandavejustice:
How are we going to collect metrics? What are we trying to ask?

6a68:
The first question we want to answer is: Will people actually use this?

The simplest way to answer the question is to measure actual usage. How about we count the number of videos watched using the viewer?

The second question is: Will people like this and recommend it to their friends?

We can get an answer to the second question by looking at the install count and churn rate that Test Pilot already measures for us.

wresuolc:
Once we determine this it should live in a metrics readme doc in the repository.  @johngruen will probably have good ideas

wresuolc:
https://github.com/meandavejustice/min-vid/blob/master/docs/metrics.md

PR #39, opened 2016-05-31
Add metrics doc

meandavejustice:
- refs #8
- add barebones metrics doc


wresuolc:
Let's land and iterate!  I'll work on a PR

PR #46, opened 2016-06-01
update metrics documentation

clouserw:
r? @meandavejustice @kaganjd


clouserw:
This is a first draft.  Comments and improvements welcome.  We're not recording enough data to answer the "Other Interesting Questions" but I'm not sure that's important.

also CC @johngruen

meandavejustice:
This looks great to me, thanks @clouserw

kaganjd:
Yeah, so cool @clouserw! 

I did have a few questions: 

1.  What does 'deactivate' mean? Is that 'x'ing the min-vid player or something else? 
2.  Are you collecting data on resize, move instances or start size/end size and start position/end position? Not sure if it matters, just wondering.

wresuolc:
1) I was assuming there was a way to deactivate.  Like, "send back to regular place on the webpage" or something?  Speaking of that, is it required to keep the original webpage open while playing the video?

2) Not collecting any data about window size or location.  We could, I was just being lazy.  Is there a standard way people record that stuff these days in a string or do we need 4 separate columns for x, y, height width?

wresuolc:
New patch adds size/location.  Still not sure about the deactivation -- @meandavejustice or @kaganjd ?

meandavejustice:
@kaganjd 
1. Deactivate in this case means should mean closing the mini player. Though in addition we should add `send-to-tab`
2. looks like wil addressed this. which looks :+1: to me.

PR #51, opened 2016-06-06
chore(eslint): Fix some ESLint warnings

pdehaan:
r? @meandavejustice @kaganjd

### DIFF
--- a/data/controls.js
+++ b/data/controls.js
@@ -4,34 +4,34 @@
  * http://mozilla.org/MPL/2.0/.
  */
 
-document.querySelector(".controls").style.display = "none";
-document.querySelector(".video-wrapper").style.display = "none";
+document.querySelector('.controls').style.display = 'none';
+document.querySelector('.video-wrapper').style.display = 'none';
 
-Array.from(document.querySelectorAll("a")).forEach(el => {
-  el.addEventListener("click", ev => {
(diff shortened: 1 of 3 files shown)


data/controls.js
```diff
@@ -4,34 +4,34 @@
  * http://mozilla.org/MPL/2.0/.
  */
 
-document.querySelector('.controls').style.display = 'none';
-document.querySelector('.video-wrapper').style.display = 'none';
+document.querySelector(".controls").style.display = "none";
```
meandavejustice: why `"` over `'`?

pdehaan: No reason at all. I think when I did the initial .eslintrc file, I tried both and we used double quotes 60% of the time and single quotes the other 40%.
I'm happy to switch everything to single quotes if you prefer. It should be a trivial fix with the ESLint `--fix` flag which will swap them all for us.

meandavejustice: yeah let's do single quotes to be consistent with testpilot, otherwise I'll get caught up on it all the time :)

pdehaan: Done!

meandavejustice:
Thank you @pdehaan

pdehaan:
We should probably do #52 (Travis-CI integration) soon, so we'll have a full time ESLint shame-bot for each commit.

PR #56, opened 2016-06-09
Add Travis-CI badge to README

pdehaan:
... also moved the README images into docs/ to unclutter the root folder a bit.

r? @meandavejustice @kaganjd


meandavejustice:
Sweet, this should kick off our travis build too so we can see if it works.

PR #58, opened 2016-06-23
Adds Vine support on Twitter and certain Vine.co URLs

kaganjd:


meandavejustice:
Couple things before this can be merged:
* remove all of the `.DS_STORE` files. [creating global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) you will probably want [sublime-text](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore) & [OSX](https://github.com/github/gitignore/blob/master/Global/OSX.gitignore)
* remove ` @min-vid-0.0.1.xpi`
* address comments in diff
* fix [linting issues](https://travis-ci.org/meandavejustice/min-vid/builds/139848449). checkout [spaces/tabs in sublime text](https://css-tricks.com/changing-spaces-tabs-sublime-text/). You can test by running `npm run lint`.

index.js
```diff
@@ -82,11 +100,47 @@ cm.Item({
...
+                 "  self.postMessage(document.URL);" +
+                 "});",
+  onMessage: function(pageURL) {
+    updatePanel(constructVineEmbedUrl(pageURL));
+  }
+});
+
+function checkLocation(opts) {
+	if (opts.documentURL.indexOf('vine.co/v/') > 0) {
+		return true;
```
meandavejustice: spacing is weird, looks like you are using tabs instead of spaces here.

index.js
```diff
@@ -82,11 +100,47 @@ cm.Item({
...
+  contentScript: "self.on('click', function (node, data) {" +
+                 "  self.postMessage(document.URL);" +
+                 "});",
+  onMessage: function(pageURL) {
+    updatePanel(constructVineEmbedUrl(pageURL));
+  }
+});
+
+function checkLocation(opts) {
+	if (opts.documentURL.indexOf('vine.co/v/') > 0) {
```
meandavejustice: These should be `> -1` as an index of `0` would still mean that it was in the url, just at the first slot.

index.js
```diff
@@ -82,11 +100,47 @@ cm.Item({
...
 });
 
+// Vine context menu on Vine.co
+// only works from direct embed URLs structured like this: 
+// https://vine.co/v/[vineID] or https://vine.co/v/[vineID]/embed
+cm.Item({
+  label: "Send to mini player",
+  context: [
+  	cm.PredicateContext(checkLocation),
+  	cm.SelectorContext('[src*="mediasource:https://vine.co/"]')
```
meandavejustice: Does this prevent this from launching with regular vine links? https://vine.co/v/hW5WxZVjJYP

index.js
```diff
@@ -82,11 +100,47 @@ cm.Item({
...
+                 "});",
+  onMessage: function(pageURL) {
+    updatePanel(constructVineEmbedUrl(pageURL));
+  }
+});
+
+function checkLocation(opts) {
+	if (opts.documentURL.indexOf('vine.co/v/') > 0) {
+		return true;
+	} else if (opts.documentURL.indexOf('twitter.com') > 0) {
```
meandavejustice: why don't we want to embed vines from twitter.com?

meandavejustice: I think we just want to check that the url is shaped like `vine.co/v/` here right? In that case we would need to check both `hrefURL` and `srcURL`

index.js
```diff
@@ -51,8 +57,20 @@ panel.port.on('link', opts => {
...
+	return (url.indexOf('youtube') > -1);
+}
+
+function isVine(url) {
+	return (url.indexOf('vine') > -1); 
+}
+
 function parseYoutubeId(src) {
-  return src.substr(src.indexOf('embed/') + 6);
+  	return src.substr(src.indexOf('embed/') + 6);
```
meandavejustice: weird spacing

index.js
```diff
@@ -51,8 +57,20 @@ panel.port.on('link', opts => {
...
+}
+
+function parseVineUrl(src) {
```
meandavejustice: can we name this something more descriptive, like `getVinePageURL`?

data/default.html
```diff
@@ -60,13 +62,16 @@
      frame.onload = updatePlayer;
 
      function updatePlayer(ev) {
-       if (!frame.src) return;
```
meandavejustice: Can you add this line back in?

data/default.html
```diff
@@ -49,7 +49,9 @@
     <script src="https://www.youtube.com/iframe_api"></script>
     <script>
      console.log('default.html loaded');
-     function onYouTubeIframeAPIReady() {console.log('yt api ready');} // for debugging
+     function onYouTubeIframeAPIReady() {
```
meandavejustice: was this for linting? or can it stay formatted to one line?

kaganjd: I put this back on one line

data/default.html
kaganjd: Yes, added back in

index.js
kaganjd: fixed my tab/space settings in Sublime

index.js
kaganjd: Yup

index.js
kaganjd: Thanks, changed this.

index.js
kaganjd: we do want to embed vines from twitter. I changed `> 0` to `> 1`

kaganjd: added checks for `hrefURL` and `srcURL`

index.js
kaganjd: seems like with PredicateContext, we can get rid of SelectorContext altogether, right? I'm not sure why the video isn't loading from that link though. I'm looking into it.

data/default.html
meandavejustice: looks like this may have been [lost in the sauce](https://www.youtube.com/watch?v=HY5tIymnvWw)

index.js
```diff
@@ -22,9 +22,15 @@ getActiveView(panel).setAttribute('backdrag', true);
 
 panel.port.on('link', opts => {
   var title = opts.title;
+  var ytTab = 'https://youtube.com/watch?v=' + parseYoutubeId(opts.src) + '&t=' + opts.time;
+  var vineTab = getVinePageURL(opts.src);
```
meandavejustice: these should be moved down into the `isYoutube` and `isVine` check since they are only used within those blocks

index.js
```diff
@@ -126,13 +126,9 @@ cm.Item({
 });
 
 function checkLocation(opts) {
```
meandavejustice: I'm not really seeing a use for this function, the only place we use it also checks that the url is a vine url. I don't know why we wouldn't want to launch min-vid from urls of that type when not on twitter or vine.com. Let's remove it and the call to is above.

kaganjd: There are two different functions for constructing Vine embed URLs and they depend on the location of the link. Twitter and certain Vine vids don't have an href in the element that we can use to construct the embed, so that's what the checkLocation thing is for. Let me know if you can think of another way to do that. I think just erasing the function will mean that we can't send to Min-Vid from certain contexts.

kaganjd: So, with checkLocation function (without checkLocation, I couldn't get these to work): 
screen shot 2016-06-27 at 1 14 51 pm
screen shot 2016-06-27 at 1 14 05 pm

meandavejustice: I see, I think we should still find another way. We should be checking based on the embed element or link itself. That way when embeds are on different sites from vine or twitter (like reddit or 4chan or tumblr or someone's wordpress blog) we won't have to add specific handling.

meandavejustice: We can work on this tomorrow together.

kaganjd: I just looked at the elements again and they're <video> which I think you said is not Vine support but HTML5 support? So maybe, for this PR, I can delete checkLocation and where it's called, push this up so at least Vine hrefs are supported, and then open a new issue for HTML5 support. What do you think?

meandavejustice: Let's leave this pr as it is for now, the changes I was working on yesterday will likely effect the vine urls code now. HTML5 Video support is encapsulated in #6 but I'm fine with have a more direct issue for it

index.js
```diff
@@ -51,10 +57,22 @@ panel.port.on('link', opts => {
   }
 });
 
+function isYoutube(url) {
```
pdehaan: Nit: `isYouTube(url)` (uppercase "T")

PR #59, opened 2016-06-28
25 move panel

kaganjd:


meandavejustice:
Thanks @kaganjd, This looks great!

PR #65, opened 2016-07-07
refactor min-vid, major improvements

meandavejustice:
- add vimeo support
- refactor ui, now uses react.js and single source of truth
at window.AppData
- switch to using get-video-id dependency
- get rid of iframe and dependencies on embed apis
- we now fetch the direct link and control a <video> element
- refs #6


meandavejustice:
@kaganjd @6a68 I've added both of you as reviewers so we can use this as an opportunity for both of you to ask as many questions as possible

PR #83, opened 2016-07-15
Tweak ESLint for React and fix warnings

pdehaan:


.eslintignore
```diff
@@ -1,3 +1,2 @@
 bootstrap.js
 data/bundle.js
-app.js
```
pdehaan: Tweaked the ESLint rules below to support JSX so it won't choke on the React elements.

.travis.yml
```diff
@@ -3,10 +3,5 @@ language: node_js
 node_js:
   - 4
 
-before_script:
-  - npm i addons-linter -g
-
 script:
-  - npm run lint
   - npm run package
-  - addons-linter dist/addon.xpi
```
pdehaan: Added **addons-linter** as a devDependency so this works locally too. Also moved the `lint` and `addons-linter` steps as npm hooks for the `package` script so running `$ npm run package` will automatically pre-lint and post-addons-linter as expected. \*magic\*

meandavejustice:
Thanks @pdehaan! I was having trouble getting eslint to recognize the react addon. It's great to have this working right!

PR #94, opened 2016-07-21
Tooltips for controls

### LINKED ISSUE #66: tooltips for controls

meandavejustice:
- added tooltips to all controls in min-vid
- added react-tooltip dependency
- fresh build of addon
- fixes #66 

post-review:
* reworked ReactTooltip to use `data-tip` to simplify code as @pdehaan suggested
* added `classnames` dep for simplifying `hidden` class logic, thanks @pdehaan 
* fixed issue where pause tooltip was displaying as "play"
* fixed weird minimized control panel css issue
* filed https://github.com/meandavejustice/min-vid/issues/110 for issue with tooltip placement when minimized
* fixed issue of tooltips being cut off by window when maximized

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,9 +1,7 @@
 const React = require('react');
 const ReactDOM = require('react-dom');
-
-// uncomment for manual testing purposes
-// const testSrc = 'https://r17---sn-ab5l6n7r.googlevideo.com/videoplayback?mv=m&mt=1468352728&ms=au&id=o-AE03DODlnfLGm9N8CcnmyLtuxZePWALwIV9O1-aQoA-H&expire=1468374597&mime=video%2Fmp4&ip=65.88.88.176&requiressl=yes&pl=23&mn=sn-ab5l6n7r&mm=31&source=youtube&lmt=1439378563427365&cnr=14&sparams=cnr%2Cdur%2Cid%2Cinitcwndbps%2Cip%2Cipbits%2Citag%2Clmt%2Cmime%2Cmm%2Cmn%2Cms%2Cmv%2Cnh%2Cpl%2Cratebypass%2Crequiressl%2Csource%2Cupn%2Cexpire&nh=IgpwcjA1LmxnYTA3KgkxMjcuMC4wLjE&sver=3&initcwndbps=1900000&signature=961892B3A4873E80CBE75F9FE5B872D2E2A8060C.7A76E7058BC532B524B2C24C06BE30C72BA527EA&dur=154.807&itag=22&fexp=9416126%2C9416891%2C9419452%2C9422596%2C9425078%2C9428398%2C9431012%2C9431452%2C9431677%2C9431838%2C9433096%2C9433380%2C9433946%2C9435036%2C9435526%2C9435876%2C9437066%2C9437553%2C9437742%2C9438280%2C9438361%2C9438661%2C9438662%2C9438734%2C9439652%2C9439965%2C9440047%2C9440302%2C9440376%2C9440542%2C9440850&key=yt6&upn=SA-YwegWsYw&ratebypass=yes&ipbits=0';
-// const vtestSrc = 'https://08-lvl3-pdl.vimeocdn.com/01/3108/0/15540231/29736282.mp4?expires=1468015395&token=0623f1dfc4b0c96f737a2';
+const ReactTooltip = require('react-tooltip');
+const cn = require('classnames');
 
 const defaultData = {
   id: '',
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -102,7 +103,8 @@ const LoadingView = React.createClass({
   render: function() {
     return (
         <div className={'loading'} onMouseEnter={this.enterView} onMouseLeave={this.leaveView}>
-          <a onClick={this.close} className={'close ' + this.state.hovered ? '': 'hidden'} />
+        <a onClick={this.close} data-tip data-for={'close-tip'} className={'close ' + this.state.hovered ? '': 'hidden'} />
```
pdehaan: In [**activity-stream**](https://github.com/mozilla/activity-stream/blob/91763bb48083be73ea910fc9e02a7771e9ef6b21/content-src/components/Loader/Loader.js#L12) we use a module named [**classnames**](http://npm.im/classnames) which may be useful (or add unnecessary bloat, who knows).

```js
const classNames = require('classnames');

<a onClick={this.close}
   data-tip
   data-for={'close-tip'}
   className={classNames('close', {hidden: this.state.hovered})} />
```

... But cuts down on a bunch of inline ifs.

pdehaan:
Play and Pause have the same tooltips.

youtube

youtube

pdehaan:
youtube
Only the "Play" and "Close" buttons have tooltips in the minimized version.

pdehaan:
This is a pretty cool tooltip module!

It looks like we may be able to hack it where we define the tooltips inline as an attribute, and then can get away with using a single `<ReactTooltip place="bottom" effect="solid" />` instance as the defaults. Of course maybe I'm completely wrong, but it Worked For Me(tm) locally.

Also removed a bit of excess binding around string attributes since it seems to work fine without it.

```js
   <div className={'controls ' + (this.state.hovered ? '' : 'hidden ') + (this.props.minimized ? 'minimized' : '')}
             onMouseEnter={this.enterControls} onMouseLeave={this.leaveControls}>
          <div className='left'>
            <ReactTooltip place='bottom' effect='solid' />

            <a onClick={this.play} data-tip='Play'
               className={this.props.playing ? 'play hidden' : 'play'} />
            <a onClick={this.pause} data-tip='Pause'
               className={this.props.playing ? 'pause' : 'pause hidden'} />
            <a onClick={this.mute} data-tip='Mute'
               className={!this.props.muted ? 'mute' : 'mute hidden'} />
            <a onClick={this.unmute} data-tip='Unmute'
               className={this.props.muted ? 'unmute' : 'unmute hidden'} />
            <input type='range' className={this.state.showVolume ? 'volume' : 'volume hidden'}
                   min='0' max='1' step='.01' value={this.props.volume} onChange={this.setVolume}/>
          </div>

          <div className='right'>
            <a onClick={this.sendToTab} data-tip='Send to tab' className='tab' />
            <a onClick={this.minimize} data-tip='Minimize'
               className={!this.props.minimized ? 'minimize' : 'minimize hidden'} />
            <a onClick={this.maximize} data-tip='Maximize'
               className={this.props.minimized ? 'maximize' : 'maximize hidden'} />
            <a onClick={this.close} data-tip='Close' className='close' />
          </div>
        </div>
```

Possibly a few other references I overlooked, so you'll probably want to do a Ctrl+F for `data-for=` locally if the above looks kosher (and isn't full of stupid Peter typos).

6a68:
This looks great overall.

I think it would be nice to eventually move to a one-tooltip model, as @pdehaan mentioned, since the user will only ever hover a single element at a time. I'd suggest we file a followup bug for that, though, rather than keep this PR open any longer.

Could the tooltip be positioned to the left or right for the controls at the corners of the player window? The 'play' and 'close' tooltips are cut off for me, because they are positioned at bottom:





Again, I think it'd be better to file followup bugs to position the tooltips off to one side. That might even be a Good First Bug for a new contributor :-)

I'm hitting some kind of error which is stopping me from testing further, but let's land this now and see what happens next. @meandavejustice should I merge, or do you want to enforce any kind of ordering in these PR merges?

6a68:
Hmm, positioning the tooltips off to one side obscures the other controls and makes it hard to slide your mouse from control to control. Definitely file a bug and we can ask UX to figure out the best available option.

meandavejustice:
* reworked ReactTooltip to use `data-tip` to simplify code as @pdehaan suggested
* added `classnames` dep for simplifying `hidden` class logic, thanks @pdehaan 
* fixed issue where pause tooltip was displaying as "play"
* fixed weird minimized control panel css issue
* filed https://github.com/meandavejustice/min-vid/issues/110 for issue with tooltip placement when minimized
* fixed issue of tooltips being cut off by window when maximized

PR #97, opened 2016-07-22
Open tab with current time of playback

### LINKED ISSUE #86: Total video timestamp is oddly decimal based
```sh
$ git rev-parse HEAD # bc4abd73a0f2cf089c147b9cfa0dbbe0d8824430
```

### Steps to reproduce:
1. Build and install add-on locally.
1. Navigate to https://www.reddit.com/r/videos and find a video from "(youtube.com)".
1. Right-click the link and select "Send to mini-player" from context menu. I tested with https://www.youtube.com/watch?v=wLww3KpLEoo
1. Play the video.
1. Note the playback length

### Actual results:
When video stops playing, the timestamp is: `2:70 / 2.70`
I have no idea what that means, but when viewing the same video in YouTube directly, it reports the time as `0:06 / 2:41`, so I'm guessing we aren't doing some Maths correctly.

---

**UPDATE:** https://m.youtube.com/watch?v=yceGT6lrn1Y shows 0:93 / 5.69

### LINKED ISSUE #90: Jump to specific time on YouTube when opening video in new tab
```sh
$ git rev-parse HEAD # bc4abd73a0f2cf089c147b9cfa0dbbe0d8824430
```

### Steps to reproduce:
1. Build and install add-on locally.
1. Navigate to https://www.reddit.com/r/videos and find a video from "(youtube.com)".
1. Right-click the link and select "Send to mini-player" from context menu. I tested using https://www.youtube.com/watch?v=yceGT6lrn1Y
1. Start watching the video for a minute.
1. Click the "Open in new tab" button to open the picture-in-picture video in a new tab.

### Actual results:
Even though I've watched 1:86 minutes of the video, when we open the video in a new tab, it always resets me back to 0:00 in the new tab.
Not sure if it'd make sense to try and open the video at the current time of the mini player.

I'm pretty sure YouTube supports this via the `t` URL param: https://youtu.be/UM-Q_zpuJGU?t=363 (which I believe is 363 seconds into a video, or in my case, starting at 6:03 into my 19:15 video clip).

Not sure if Vimeo supports something like that (I'll investigate shortly) and probably doesn't make sense for smaller Vine videos or some other services with shorter clips.

meandavejustice:
- fixes #90
- also fixed formatting of 'currentTime / duration' in player, fixes #86

### DIFF
--- a/app.js
+++ b/app.js
@@ -53,9 +53,12 @@ window.AppData = new Proxy(defaultData, {
 //   });
 // }, 8000);
 
-function formatTime(c, d) {
-  // example:   0:35 / 2:33
-  return ((c / 60).toFixed(2) + ' / ' + (d / 60).toFixed(2)).replace('.', ':');
+function formatTime(seconds) {
+  const now = new Date(seconds * 1000);
+  const utc = new Date(now.getTime() + (now.getTimezoneOffset() * 60 * 1000));
+  return utc.toLocaleTimeString('en-US', {hour12: false})
+    .replace(/^00:/, '') // Strip leading "00:" if hours is empty.
+    .replace(/^0/, '');  // Strip leading "0" in minutes, if any.
 }
 
 const App = React.createClass({
@@ -122,7 +125,7 @@ const PlayerView = React.createClass({
   },
   step: function() {
     window.AppData = Object.assign(window.AppData, {
-      currentTime: formatTime(this.refs.video.currentTime, window.AppData.duration),
+      currentTime: `${formatTime(this.refs.video.currentTime)} / ${formatTime(window.AppData.duration)}`,
       progress: this.refs.video.currentTime / window.AppData.duration
     });
 
@@ -190,7 +193,8 @@ const PlayerView = React.createClass({
     sendToAddon({
       action: 'send-to-tab',
       id: window.AppData.id,
-      domain: window.AppData.domain
+      domain: window.AppData.domain,
+      time: this.refs.video.currentTime
     });
   },
   close: function() {
--- a/index.js
+++ b/index.js
@@ -26,7 +26,7 @@ panel.port.on('message', opts => {
   const title = opts.action;
 
   if (title === 'send-to-tab') {
-    const pageUrl = getPageUrl(opts.domain, opts.id);
+    const pageUrl = getPageUrl(opts.domain, opts.id, opts.time);
     if (pageUrl) require('sdk/tabs').open(pageUrl);
     else console.error('could not parse page url for ', opts); // eslint-disable-line no-console
     panel.hide();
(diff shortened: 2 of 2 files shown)


data/bundle.js
```diff
@@ -54,9 +54,18 @@ window.AppData = new Proxy(defaultData, {
...
+}
+
 function formatTime(c, d) {
   // example:   0:35 / 2:33
-  return ((c / 60).toFixed(2) + ' / ' + (d / 60).toFixed(2)).replace('.', ':');
+  const cMin = Math.floor(c / 60);
+  const cSec = Math.floor(c - cMin * 60);
+  const dMin = Math.floor(d / 60);
+  const dSec = Math.floor(d - dMin * 60);
+  return `${ cMin }:${ padNum(cSec) } / ${ dMin }:${ padNum(dSec) }`;
```
pdehaan: Just to confirm, this is a generated file, right?

meandavejustice: yeah, you can ignore it for reviews. It will be ignored by git once we start hosting the addon somewhere else

app.js
```diff
@@ -53,9 +53,18 @@ window.AppData = new Proxy(defaultData, {
...
+}
+
 function formatTime(c, d) {
   // example:   0:35 / 2:33
-  return ((c / 60).toFixed(2) + ' / ' + (d / 60).toFixed(2)).replace('.', ':');
+  const cMin = Math.floor(c / 60);
+  const cSec = Math.floor(c - cMin * 60);
+  const dMin = Math.floor(d / 60);
+  const dSec = Math.floor(d - dMin * 60);
+  return `${cMin}:${padNum(cSec)} / ${dMin}:${padNum(dSec)}`;
```
pdehaan: Not sure if you want to maybe break this down one more level since it looks pretty duplicated.

```js
function formatTime(current, total) {
  const padNum = (num) => {
    return ("00" + num).substr(-2);
  };
  // example:   0:35 / 2:33
  const duration = (time) => {
    const minutes = Math.floor(time / 60);
    const seconds = Math.floor(time - (minutes * 60));
    return `${ minutes }:${ padNum(seconds) }`;
  };
  return `${ duration(current) } / ${ duration(total) }`;
}

console.log(formatTime(6543, 7387)); // Output: 109:03 / 123:07
```

My variable names are kind of :poop:, but hey. Plus we also may want to test it with longer videos, like https://www.youtube.com/watch?v=HflDxjLHDNU (which is 2:03:07). Not sure if that would display as 123:07, or if we want to add more logic for potential hour long videos so it displays as 2:03:07.

meandavejustice: Nice, good thinking with the `> an hour` check

pdehaan: It's a bit edge case, but YouTube does have a few full length movies, or other tech-like videos that exceed 60 minutes.
There may be better ways to format timestamps, but possibly not worth the complexity of creating a new Date object and then trying to use Date and then handle some grosser logic:

```js
function formatTime(seconds) {
  const timestamp = new Date(seconds * 1000);
  const utcTimestamp = new Date(
    timestamp.getUTCFullYear(),
    timestamp.getUTCMonth(),
    timestamp.getUTCDate(),
    timestamp.getUTCHours(),
    timestamp.getUTCMinutes(),
    timestamp.getUTCSeconds()
  );
  return utcTimestamp.toLocaleTimeString('en-US', {
    hour12: false,
    hour: '2-digit',
    minute: '2-digit',
    second: '2-digit'
  }).replace(/^0+:?/, '');
}

console.log(formatTime(7384)); // "2:03:04"
console.log(formatTime(654));  // "10:54"
console.log(formatTime(99));   // "01:39"
console.log(formatTime(9));    // "00:09"
```

meandavejustice: I like this solution best so far. Seems to be the most _correct_

meandavejustice: Thanks again for this @pdehaan, you rule!

6a68:
@meandavejustice R+ merge at will

PR #93, opened 2016-07-21
Overlay icon

meandavejustice:
- fixes #64 
- add overlay-icon.js for adding icon to dom and handling actions
- add overlay-icon.css for styling icon
- supports youtube(watch/home) and vimeo(detail/home) as well as vimeo and youtube embeds
- fix orphaned audio streams on `send-to-tab` and `close` fixes #84 

I've tried to find a good percentage ratio for the `width`, `top`, and `left` css attributes for the icon container. Something that will look reasonable for the many different sizes. I'm happy to take suggestions for tweaks to this, but I'd like to not hold up this pr for css polish that can be covered in #72 

### Links for testing:
- [youtube home](https://www.youtube.com)
- [youtube watch](https://www.youtube.com/watch?v=Wjku6BZz3EQ)
- [vimeo home](https://vimeo.com/home)
- [vimeo detail](https://vimeo.com/15540231)
- [vimeo embed](http://www.ubu.com/film/anger_puce.html)

### Testing plan:
1. Install addon
2. test both "youtube watch" and "vimeo detail" links above with the context menu
3. navigate to all links above and make sure videos can be sent to min-vid via overlay icon.

### Known issues:
- launching min-vid from `vimeo detail` or `vimeo embed` also starts the video. It seems `stopImmediatePropagation` and `addEventListener capture argument` aren't working. Will probably file a separate issue to be fixed later.
- There seems to be an issue getting the icons attached when navigating around youtube.com, this seems to be because it is a single page app, and [page-mod](https://duckduckgo.com/l/?kh=-1&uddg=https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FMozilla%2FAdd-ons%2FSDK%2FHigh-Level_APIs%2Fpage-mod) only runs on new pages. This will probably be a simple fix by just running the attach methods every once in a while. I'm holding out for a bit to see if we can think of a more elegant way than an interval.

PR #104, opened 2016-07-27
Add ESLint curly rule

pdehaan:
Ref: https://github.com/meandavejustice/min-vid/pull/93#discussion_r72361992


6a68:
@meandavejustice Your call on syntax rules

meandavejustice:
I like the ability to write if statements on one line. I think there are cases where it can be more understandable. I don't like being bound in this way. thanks for opening @pdehaan but I'm going to pass on adding this rule.

PR #115, opened 2016-08-03
WIP metrics

6a68:
@meandavejustice Hey, does this seem like a good approach? When state gets updated, call a metrics function with the name of the property, old value, and new value; if the changed property seems interesting, fire a ping.

Currently handles the user loading a video or clicking the controls (`play`, `pause`, `minimize`, `maximize`, `send to tab`, `close`).

I skipped `volume` because moving a slider fires a ton of updates, and skipped `progress` because it fires constantly. I think `activate` and `deactivate` should be reported by the test pilot addon, but I need to double check. Still need to handle `end` and `available`, and that should wrap up the list in issue 48.

### DIFF
--- a/app.js
+++ b/app.js
@@ -20,11 +20,53 @@ const defaultData = {
 
 window.AppData = new Proxy(defaultData, {
   set: function(obj, prop, value) {
+    // TODO: if more than one non-UI component needs to detect state changes,
+    // update this function to use proper pub-sub: subscribe to state changes,
+    // call callbacks on matching state change (or find a nice library).
+    // Inlining metrics code for now.
+    if (prop in obj && obj[prop] !== value) {
+      console.log(Date.now() + ': changing ' + prop + ' from ' + obj[prop] + ' to ' + value);
+      recordEvent({
+        state: prop,
+        oldValue: prop in obj && obj[prop] || null,
+        newValue: value || null
+      });
+    }
     obj[prop] = value;
     renderApp();
     return true;
   }
 });
+
+// recordEvent filters through state changes and sends interesting events to
+// telemetry, along with the app's current state.
+//
+// o is an object with the keys state, oldValue, newValue.
+function recordEvent(o) {
+  // Always ignore 'progress' changes, because they fire a LOT.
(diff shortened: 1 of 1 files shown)


meandavejustice:
@6a68 Yeah, I think this is a good approach. Nice job

app.js
```diff
@@ -20,12 +20,54 @@ const defaultData = {
 
 window.AppData = new Proxy(defaultData, {
   set: function(obj, prop, value) {
+    // TODO: if more than one non-UI component needs to detect state changes,
+    // update this function to use proper pub-sub: subscribe to state changes,
+    // call callbacks on matching state change (or find a nice library).
+    // Inlining metrics code for now.
+    if (prop in obj && obj[prop] !== value) {
+      console.log(Date.now() + ': changing ' + prop + ' from ' + obj[prop] + ' to ' + value);
```
pdehaan:
```
28:7  error  Unexpected console statement  no-console
```

We need to add an `// eslint-disable-line no-console` to this line.

app.js
```diff
@@ -20,12 +20,54 @@ const defaultData = {
...
+  // state: 'close', oldValue: false, newValue: true
+  // state: 'send-to-tab', oldValue: false, newValue: true
+  // TODO: will this make sense on the server / dashboard?
+  //
+  // Note: ignoring the volume and progress sliders, which change continuously.
+  // We can later add custom events like we do for 'close' and 'send-to-tab'.
+  const loggedEvents = ['playing', 'minimized', 'muted', 'close', 'send-to-tab'];
+  if ((o.state === 'id' && !!o.newValue) || loggedEvents.indexOf(o.state) > -1) {
+    // Just send all the things and figure out the details on the server.
+    sendPing(o, window.AppData);
```
pdehaan:
```
67:5  error  'sendPing' is not defined     no-undef
```

 Not sure where `sendPing()` is defined, but ESLint seems unhappy.

6a68: Yeah, it's not defined yet, but I'll fill it in before submitting for a real review :-)

6a68:
Closing, thanks for the comments all

PR #122, opened 2016-08-09
Fix #121

### LINKED ISSUE #121: Add support for Vines from Vine.co homepage
We get the video URL from Vine embeds by loading the video on its single-video page on vine.co, then looking at the `twitter:player:stream` meta tag in the page header.

Unfortunately, on the vine.co homepage, videos are embedded as a preview jpeg, and when the user mouses over the video, some kind of Ember video player UI is overlaid on the preview image. The problem is that this video player is a single component shared by the whole vine.co homepage, so a right click on the playing image won't point to a useful part of the DOM. To make matters worse, after right-clicking and while mousing down to 'send to mini player', Ember detects the mouse movement and empties out the video player--so you wind up with a right click event on an empty element.

If we could somehow get around the Ember component, it looks like you can get from the video preview image to the vine ID as follows:
- `.video-container` contains the preview image), but its next sibling, `.info`, contains a couple of descendants that lead to the vine ID, `.post-metadata .time a`:
```
<div class="post-metadata">
  <span class="time">
    <a href="/v/5bzXw9zMQ0n" class="ember-view" id="ember1439">2h ago</a></span>
  </span>
```  

Maybe we could use the right-click's page coords to infer what element was actually right-clicked? That would let us get to the correct Vine card in the DOM.

kaganjd:
- Add support for Vines from Vine.co homepage
- Change contentScript to send the poster instead of the href
- Parse the contents of poster to get the mp4 URL

### DIFF
--- a/index.js
+++ b/index.js
@@ -126,6 +126,21 @@ cm.Item({
   }
 });
 
+cm.Item({
+  label: contextMenuLabel,
+  context: [
+    cm.URLContext(['https://vine.co/*']),
+    cm.SelectorContext('video')
+  ],
+  contentScript: 'self.on("click", function (node, data) {' +
+              ' self.postMessage(node.poster);' +
+              ' });',
+  onMessage: function(url) {
+    const mp4 = url.replace(/thumbs/, 'videos').split('.jpg')[0];
+    updatePanel({src: mp4});
+  }
+});
+
 function updatePanel(opts) {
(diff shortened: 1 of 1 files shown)


meandavejustice:
Pulling this down for a manual test this afternoon.

meandavejustice:
Thanks @kaganjd

PR #124, opened 2016-08-10
Support video previews from Google search results pages

### LINKED ISSUE #114: Add support from Google search page
When I try to send YouTube vids to panel from Google, I get an endlessly loading panel.
screen shot 2016-08-01 at 12 26 33 pm

Error is: "qs.parse(...).url_encoded_fmt_stream_map is undefined"
Referencing [line 12 of get-youtube-url.js](https://github.com/meandavejustice/min-vid/blob/master/lib/get-youtube-url.js#L12).

kaganjd:
- add new URLContext and SelectorContext
- update other SelectorContexts
- add function to decode platform URLs
- getUrlFn based on decoded URL
- fixes #114

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,33 @@ const makePanelTransparent = require('./lib/make-panel-transparent.js');
 const getDocumentDimensions = require('./lib/get-document-dimensions.js');
 const pageMod = require('sdk/page-mod');
 const cm = require('sdk/context-menu');
+
+const URLS = {
+  'vimeo': ['vimeo.com/'],
+  'youtube': ['youtube.com/', 'youtu.be/'],
+  'vine': ['vine.co/']
+}
+
+// Given a video service name from the URLS object, return an href *= selector
+// for the corresponding urls.
+// Arguments:
+//   videoService: a key from URLS or '*' for all domains from URLS
+//   shouldEncode: (optional) encode domains if true
+function getSelectors(videoService, shouldEncode) {
+  let domains = [];
+  if (videoService in URLS) {
+    domains = URLS[videoService]
+  } else if (videoService === '*') {
+    domains = Object.keys(URLS).map(name => URLS[name])
+                               .reduce((prev, curr) => prev.concat(curr))
+  } else {
+    console.error(`Error: ${videoService} missing or not supported`) // eslint-disable-line no-console
+  }
+  const selectors = domains.map(url => `[href*="${shouldEncode ? encodeURIComponent(url) : url}"]`)
+                           .reduce((prev, curr) => `${prev}, ${curr}`)
+  return selectors;
+}
+
 const contextMenuLabel = 'Send to mini player';
 const contextMenuContentScript = `
 self.on('click', function (node, data) {
@@ -130,7 +157,7 @@ function getPageUrl(domain, id, time) {
 
 cm.Item({
   label: contextMenuLabel,
-  context: cm.SelectorContext('[href*="youtube.com"], [href*="youtu.be"]'),
+  context: cm.SelectorContext(getSelectors('youtube')),
   contentScript: contextMenuContentScript,
   onMessage: (url) => {
     sendMetricsData({changed: 'activate', domain: 'youtube.com'});
@@ -157,7 +184,7 @@ cm.Item({
 
 cm.Item({
   label: contextMenuLabel,
-  context: cm.SelectorContext('[href*="vimeo.com"]'),
+  context: cm.SelectorContext(getSelectors('vimeo')),
   contentScript: contextMenuContentScript,
   onMessage: (url)=> {
     sendMetricsData({changed: 'activate', domain: 'vimeo.com'});
@@ -169,7 +196,7 @@ cm.Item({
 
 cm.Item({
   label: contextMenuLabel,
-  context: cm.SelectorContext('[href*="vine.co/v/"]'),
+  context: cm.SelectorContext(getSelectors('vine')),
   contentScript: contextMenuContentScript,
   onMessage: function(url) {
     sendMetricsData({changed: 'activate', domain: 'vine.co'});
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -126,6 +126,35 @@ cm.Item({
   }
 });
 
+cm.Item({
+  label: contextMenuLabel,
+  context: [
+    cm.URLContext(['*.google.com']),
+    //TODO?: create a variable with all supported hrefs and reference it in this SelectorContext
```
meandavejustice: I agree, this should live in a variable

index.js
```diff
@@ -126,6 +126,35 @@ cm.Item({
...
+  contentScript: contextMenuContentScript,
+  onMessage: function(url) {
+    const regex = /url=(https[^;]*)/;
+    const match = regex.exec(url)[1];
+    const decoded = decodeURIComponent(match).split('&usg')[0];
+    let getUrlFn;
+    if (decoded.indexOf('youtube.com' || 'youtu.be') > -1) {
+        getUrlFn = getYouTubeUrl;
+        domain = 'youtube.com';
+      } else if (decoded.indexOf('vimeo.com')  > -1) {
```
meandavejustice: spacing/tabbing is messed up on these `} else if` lines

index.js
```diff
@@ -126,6 +126,35 @@ cm.Item({
...
+        getUrlFn = getVimeoUrl;
+        domain = 'vimeo.com';
+      } else if (decoded.indexOf('vine.co') > -1) {
+        getUrlFn = getVineUrl;
+        domain = 'vine.co';
+      }
+    launchVideo({url: decoded,
```
meandavejustice: A weird edgecase, but we should check if `domain` and `getUrlFn` are defined here before calling `launchVideo`

meandavejustice: `else` we can `console.error` for now, down the road we will show something in the ErrorView

meandavejustice:
Needs a rebase/squash

index.js
```diff
@@ -126,6 +126,35 @@ cm.Item({
...
+  ],
+  contentScript: contextMenuContentScript,
+  onMessage: function(url) {
+    const regex = /url=(https[^;]*)/;
+    const match = regex.exec(url)[1];
+    const decoded = decodeURIComponent(match).split('&usg')[0];
+    let getUrlFn;
+    if (decoded.indexOf('youtube.com' || 'youtu.be') > -1) {
+        getUrlFn = getYouTubeUrl;
+        domain = 'youtube.com';
```
meandavejustice: Need to define domain above with `let`

index.js
kaganjd: Something like `const supportedGoogleSelectors = '[href*="youtube.com%2F"], [href*="youtu.be%2F"], [href*="vine.co%2F"], [href*="vimeo.com%2F"]'` at the top?

meandavejustice: yeah, that works :)

kaganjd:
@meandavejustice I think this is done, can you take a look?

index.js
```diff
@@ -16,6 +16,32 @@ const getVineUrl = require('./lib/get-vine-url.js');
...
+  if (videoService in URLS) {
+    domains = URLS[videoService]
+  } else if (videoService === '*') {
+    domains = Object.keys(URLS).map(name => URLS[name])
+                               .reduce((prev, curr) => prev.concat(curr))
+  } else {
+    console.error('Error: Video service missing or not supported') // eslint-disable-line no-console
+  }
+  const selectors = domains.map(url => '[href*="' + (shouldEncode ? encodeURIComponent(url) : url) + '"]')
+                .reduce((prev, curr) => prev + ', ' + curr)
```
meandavejustice: We should use template strings here instead of regular string concatenation

index.js
```diff
@@ -16,6 +16,32 @@ const getVineUrl = require('./lib/get-vine-url.js');
...
+//   shouldEncode: (optional) encode domains if true
+function getSelectors(videoService, shouldEncode) {
+  let domains = [];
+  if (videoService in URLS) {
+    domains = URLS[videoService]
+  } else if (videoService === '*') {
+    domains = Object.keys(URLS).map(name => URLS[name])
+                               .reduce((prev, curr) => prev.concat(curr))
+  } else {
+    console.error('Error: Video service missing or not supported') // eslint-disable-line no-console
```
meandavejustice: Let's log the `videoService` argument here

index.js
meandavejustice: also fix the tab spacing for line 41 to match up with the `.map` call above

meandavejustice:
Looks like this needs one more rebase. Ping me when it's up to date and I'll review it right away so we can get this merged.

index.js
```diff
@@ -17,6 +17,33 @@ const makePanelTransparent = require('./lib/make-panel-transparent.js');
...
+  let domains = [];
+  if (videoService in URLS) {
+    domains = URLS[videoService]
+  } else if (videoService === '*') {
+    domains = Object.keys(URLS).map(name => URLS[name])
+                               .reduce((prev, curr) => prev.concat(curr))
+  } else {
+    console.error(`Error: ${videoService} missing or not supported`) // eslint-disable-line no-console
+  }
+  const selectors = domains.map(url => '[href*="' + (shouldEncode ? encodeURIComponent(url) : url) + '"]')
```
meandavejustice:
``` js
`[href*="${shouldEncode ? encodeURIComponent(url) : url}"]`
```

index.js
```diff
@@ -196,6 +223,35 @@ cm.Item({
...
 
+cm.Item({
+  label: contextMenuLabel,
+  context: [
+    cm.URLContext(['*.google.com']),
+    cm.SelectorContext(getSelectors('*', true)),
+  ],
+  contentScript: contextMenuContentScript,
+  onMessage: function(url) {
+    const regex = /url=(https[^;]*)/.exec(url)[1];
```
meandavejustice: We need to add a `?` after the `s` in https so that we can accept http youtube urls as well. google search sometimes embeds youtube videos in the results and those links lead with `http`
``` js
const regex = /url=(https?[^;]*)/.exec(url)[1];
```

PR #126, opened 2016-08-17
Change getVineUrl argument to take in a Vine ID

### LINKED ISSUE #120: Support for Vine.co hrefs broke
I think the `launchVideo` function that was added in this commit (https://github.com/meandavejustice/min-vid/pull/93/files#diff-168726dbe96b3ce427e7fedce31bb0bcR123) doesn't work for Vine vids because getVineUrl takes in a url, whereas the other get[Platform]Url functions take in an ID and then construct a url.

It seems like we can either update the getVineUrl function or replace the changes that were made with the old code: https://github.com/meandavejustice/min-vid/pull/93/files#diff-168726dbe96b3ce427e7fedce31bb0bcL130

kaganjd:
Fixes #120.

### DIFF
--- a/index.js
+++ b/index.js
@@ -189,8 +189,10 @@ cm.Item({
               ' self.postMessage(node.poster);' +
               ' });',
   onMessage: function(url) {
-    const mp4 = url.replace(/thumbs/, 'videos').split('.jpg')[0];
-    updatePanel({src: mp4});
+    const mp4 = url.replace(/thumbs/, 'videos').split(/\.jpg/)[0];
+    launchVideo({url: url,
+                domain: 'vine.co',
+                src: mp4});
   }
 });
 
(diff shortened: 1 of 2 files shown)


kaganjd:
@meandavejustice I think this one's done too. Review plz?

index.js
```diff
@@ -182,11 +184,15 @@ function updatePanel(opts) {
...
-  // opts {url: url, getUrlFn: getYouTubeUrl, domain: 'youtube.com'}
+  // opts {url: url,
+  //       getUrlFn: getYouTubeUrl,
+  //       domain: 'youtube.com',
+  //       src: streamURL or ''}
   const id = getVideoId(opts.url);
-  updatePanel({domain: opts.domain, id: id, src: ''});
-  opts.getUrlFn(id, function(err, streamUrl) {
+  updatePanel({domain: opts.domain, id: id, src: opts.src || ''});
+  opts.src ? updatePanel({src: opts.src}) : opts.getUrlFn(id, function(err, streamUrl) {
```
meandavejustice: This can be rewritten as:
```
if (!opts.src) {
  opts.getUrlFn(id, function(err, streamUrl) {
    if (!err) updatePanel({src: streamUrl});
  });
}
```

since on the line above calls `updatePanel` with `opts.src` we don't need to do it here.

PR #132, opened 2016-08-17
Dragging

### LINKED ISSUE #2: Make video view draggable
We need to find a component ([like panel](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/panel)) that is draggable.

meandavejustice:
- fixes #2
- cleanups. react components are now in `components/` and common utils moved to `lib/`
- adds `react-draggable@2.2.1` dependency


meandavejustice:
video from @6a68 https://www.dropbox.com/s/dod7wf0bu6spoi6/draggable-1.mov?dl=0

researching a way to get rid of the shadow on mac

index.js
```diff
@@ -189,6 +206,9 @@ function launchVideo(opts) {
   opts.getUrlFn(id, function(err, streamUrl) {
     if (!err) updatePanel({src: streamUrl});
   });
+
+  // todo: see if we can just call this when initializing the panel
+  makePanelTransparent(panel);
```
6a68: @meandavejustice Yeah, it's possible to just call this on startup, I'll file a bug

6a68:
:tada:

PR #146, opened 2016-08-22
Audit package.json and bundling scripts

meandavejustice:
- added `npm run dev` to ease local development
- added dev-dependency `http-server@0.9.0`
- moved `classnames` and `react-tooltip` deps to dev-dependencies
- added dev-dependency `pre-commit@1.1.3`
- `precommit` runs "build-script" and "package" commands, which in turn
also runs linting("prepackage" and "postpackage")
- updated `.jpmignore` to exclude excess files (license, readme) from `get-video-id` dep
- created `client-lib` dir and moved `send-to-addon.js` and `send-metrics-event.js`. This
is to avoid bloat in the xpi. `client-lib` is included in bundle and does not need to be
included in xpi.


meandavejustice:
cc/ @pdehaan @6a68

PR #157, opened 2016-08-30
Tweak CSS of overlay icon and progress bar

### LINKED ISSUE #128: CSS tweak for overlay icon?
- on hover, top corners aren't rounded and bottom background dimensions are off

**vimeo, large vid on homepage**

screen shot 2016-08-17 at 10 51 09 am

**youtube, small vid on homepage**
screen shot 2016-08-19 at 1 23 01 pm

kaganjd:
Fixes #128.

### DIFF
--- a/data/panel.css
+++ b/data/panel.css
@@ -220,6 +220,7 @@ input[type=range]::-moz-focus-outer {
     height: 50px;
     padding-top: 6px;  
     position: absolute;
+    width: 100%;
 }
 
 .progress span {
@@ -229,7 +230,6 @@ input[type=range]::-moz-focus-outer {
 
 .video-progress {
     background: #ffffffaa;
-    border-radius: 20px;
     border: none;
     height: 10px;
     margin-top: 6px;
@@ -239,7 +239,6 @@ input[type=range]::-moz-focus-outer {
 .video-progress[value]::-moz-progress-bar {
     appearance: none;
     background: #4789d7;
-    border-radius: 20px;
     border: none;
 }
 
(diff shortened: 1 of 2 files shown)


data/panel.css
```diff
@@ -229,7 +230,6 @@ input[type=range]::-moz-focus-outer {
 
 .video-progress {
     background: #ffffffaa;
-    border-radius: 20px;
```
meandavejustice: why did you remove border-radius here and below in `.video-progress[value]::-moz-progress-bar`?

kaganjd: @6a68 went back and forth about different ways to get the `[value]` to be the same shape as the `.video-progress` bar at < 2 seconds. I looked at svg masks, the background-clip css property, and few other things, but eventually it seemed to be a lot of work for such a small thing. I also wasn't sure how a mask would scale as the player gets resized. So we decided that the way to most consistently get the [value] bar to be the right shape was just to get rid of the border-radius.

kaganjd: Here's what it looks like with padding-left. A little better, but still not quite right.
screen shot 2016-08-29 at 4 47 28 pm

kaganjd:
This PR also doesn't address the `overlay-icon` interfering with some Vimeo and YT elements because changing the location of the icon on the page seems like something that warrants more UI discussion with @johngruen. I'll create a separate issue for that with the screenshots from #128

meandavejustice:
The border radius changes here seem like a regression to me. @kaganjd Would you mind discussing this further? Either here on github or in irc

PR #162, opened 2016-08-31
addon signing script

meandavejustice:
- refs #13

### DIFF
--- a/bin/sign
+++ b/bin/sign
@@ -1,0 +1,97 @@
+#!/usr/bin/env node
+/*
+ * This Source Code is subject to the terms of the Mozilla Public License
+ * version 2.0 (the 'License'). You can obtain a copy of the License at
+ * http://mozilla.org/MPL/2.0/.
+ */
+/*
+ * This script is for signing the Min Vid add-on. If the current
+ * package version has already been signed, we download the signed xpi
+ * from amo, and move the files into the dist/ directory.
+ * If the package version has not been signed, sign the xpi, and move
+ * them into the /dist directory.
+ *
+ */
+var fs = require('fs');
+var path = require('path');
+var request = require('request');
+var jwt = require('jsonwebtoken');
+var manifest = require('../package.json');
+var version = manifest.version;
+var apiKey = process.env['AMO_USER'];
+var apiSecret = process.env['AMO_SECRET'];
+var authToken = jwt.sign({iss: apiKey}, apiSecret, {
+  algorithm: "HS256",
+  expiresIn: 60
+});
+var signedOpts = {
+  url: 'https://addons.mozilla.org/api/v3/addons/@' + manifest.name + '/versions/'+version+ '/',
+  headers: {
+    'Authorization': 'JWT ' + authToken
+  }
+};
+
+request(signedOpts, signCb);
+
+function signCb(err, resp, body) {
+  if (!err && resp.statusCode == 200) {
+    var info = JSON.parse(body);
+    if (info.files.length) {
+      var ws = fs.createWriteStream('dist/addon-signed.xpi').on('finish', removeGeneratedXpi);
+      signedOpts.url = info.files[0].download_url;
+      request(signedOpts).pipe(ws);
+    }
+  } else distAddon();
+}
+
+// if we need to sign and distribute our add-on, we want to use this method
+function distAddon() {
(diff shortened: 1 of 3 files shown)


bin/sign
```diff
@@ -0,0 +1,97 @@
...
+ * If the package version has not been signed, sign the xpi, and move
+ * them into the /dist directory.
+ *
+ */
+var fs = require('fs');
+var path = require('path');
+var request = require('request');
+var jwt = require('jsonwebtoken');
+var manifestFilename = __dirname + '/../package.json';
+var manifest = require(manifestFilename);
```
pdehaan: you can probably just do `var manifest = require('../package.json');` here instead.

bin/sign
```diff
@@ -0,0 +1,97 @@
...
+
+function signAddon(xpiPath, cb) {
+  require('jpm/lib/sign').sign({
+    apiKey: apiKey,
+    apiSecret: apiSecret,
+    xpi: xpiPath
+  }).then(function(result) {
+    if (result.success) cb(null, result.downloadedFiles[0]);
+  }).catch(cb);
+}
```
pdehaan: I forget if we should use `require('jpm/lib/sign')` directly or just use the [**sign-addon**](https://github.com/mozilla/sign-addon) module to sign XPIs (or if it even matters).
/cc @kumar303

bin/sign
```diff
@@ -0,0 +1,97 @@
...
+  });
+}
+
+function signAddon(xpiPath, cb) {
+  require('jpm/lib/sign').sign({
+    apiKey: apiKey,
+    apiSecret: apiSecret,
+    xpi: xpiPath
+  }).then(function(result) {
+    if (result.success) cb(null, result.downloadedFiles[0]);
```
pdehaan: Possibly a gotcha here if the `result.success` is falsey, but nothing throws an error to the `.catch()` handler. Looks like the `cb()` callback would never get called.

bin/sign
```diff
@@ -0,0 +1,97 @@
...
+
+function checkExistsAndMv(fromFilePath, toFilePath, cb) {
+  fs.stat(fromFilePath, function(err) {
+    if (err) return cb(err);
+    fs.rename(fromFilePath, toFilePath, function(err) {
+      if (err) return cb(err);
+      else cb();
+    });
+  });
+}
```
pdehaan: Since this file is doing a lot of file operations, not sure if it'd be easier to use something like [**fs-extra**](https://github.com/jprichardson/node-fs-extra) and synchonous methods.

package.json
```diff
@@ -46,6 +47,7 @@
     "react-dom": "15.1.0",
     "react-draggable": "2.2.1",
     "react-tooltip": "3.0.13",
+    "sign-addon": "0.1.3",
```
pdehaan: Not sure if we're using this dependency anywhere [yet]. Only other reference to it that I see in this PR is my stupid comment above: https://github.com/meandavejustice/min-vid/pull/162/files#r77084681

bin/sign
meandavejustice: They appear to do the same, we already require jpm so i'd rather avoid an extra dep

package.json
meandavejustice: Whoops, meant to remove this, nice catch

bin/sign
pdehaan: Sure. I think I was just remembering this bug: https://github.com/mozilla/testpilot/issues/822
But looking closer, it looks like **jpm** switched over to **sign-addon** in [`jpm/lib/sign`](https://github.com/mozilla-jetpack/jpm/blob/413022d1c647565d24090061c1c5898ae092d24d/lib/sign.js#L11) anyways, so I'm just going to shut up meow.

kumar303: Since `jpm` is technically deprecated (in favor of [web-ext](https://github.com/mozilla/web-ext)) then I'd suggest using `sign-addon` directly. It will be updated more frequently. However, `jpm` isn't going away anytime soon.

bin/sign
meandavejustice: added else clause

bin/sign
meandavejustice: gunna put this off for now, I just grabbed most of this script from the TxP repo

PR #168, opened 2016-09-01
add update URL

clouserw:
Once this lands, can you give me a new signed add-on and then I can upload that to s3?

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "repository": "meandavejustice/min-vid",
   "main": "index.js",
   "author": "meandave",
+  "updateURL": "https://testpilot.firefox.com/files/min-vid/updates.json",
   "engines": {
     "firefox": ">=38.0a1"
   },


package.json
```diff
@@ -8,6 +8,7 @@
   "repository": "meandavejustice/min-vid",
   "main": "index.js",
   "author": "meandave",
+  "updateURL": "https://testpilot.firefox.com/files/min-vid/updates.json",
```
meandavejustice: `updateURL` needs to be `path/to/update.rdf`
[example from TxP addon](https://github.com/mozilla/testpilot/blob/master/addon/package.json#L19)

meandavejustice:
closing for #171

package.json
wresuolc: rdf is old and busted.  .json is new and hot.  Also, please get rid of the updateLink you added - that'll come from the json (and I'll add a checksum to it).

wresuolc:
merging this one doesn't matter, but reopening so we don't lose track of it

meandavejustice:
closing for #174 

- fixed updateURL to point to `updates.json`
- remove call to update-version bin script
- remove handling of `update.rdf` from package.json
- bump version

PR #175, opened 2016-09-02
rename min-vid to Min Vid 

clouserw:


package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "title": "min-vid",
+  "title": "Min Vid",
   "name": "min-vid",
-  "version": "0.1.4",
+  "version": "0.1.5",
```
clouserw: I only updated this so we don't forget later on.  I don't think we need to roll a new version just for the name change.

meandavejustice:
Thanks @clouserw !

PR #182, opened 2016-09-02
Run ESLint on bin/ files

pdehaan:
Noticed that the files in bin/* don't have any extensions, so therefore don't get automatically linted by ESLint.

This PR now lints everything in bin/*, and auto-fixed all the issues using ESLint's awsome `--fix` flag, via `$ npm run lint -- --fix`.
No interesting errors. Mostly just changing`const` to `let`/`var`, one `===` violation, removed a couple unused deps, and me just disabling the `no-console` rule for the CLI tools.

I don't have/need AMO creds to fully test this, so @meandavejustice may want to give it a test drive before merging to make sure I didn't Peter something up horribly.

### DIFF
--- a/bin/sign
+++ b/bin/sign
@@ -12,20 +12,22 @@
  * them into the /dist directory.
  *
  */
-var fs = require('fs');
-var path = require('path');
-var request = require('request');
-var jwt = require('jsonwebtoken');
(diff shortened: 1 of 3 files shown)


6a68:
hey, let's hit the big green button and see what happens

PR #193, opened 2016-09-05
Draggable sdk panel cleanups

meandavejustice:
- remove `react-draggable` dep
- move `draggifyPanel` method to `lib/make-panel-draggable.js`
- remove `lib/get-document-dimensions.js`
- remove handling of `expand-panel` and `shrink-panel`
- remove draggable code from `components/app-view.js`
- fixes #136 
- fixes #167
- fixes #155
- fixes #154
- fixes #145
- fixes #149
- fixes #144
- fixes #140
- fixes #197

Issue #195, opened 2016-09-06
The "Click here to move the thing" bar should change its position according to the player's position on the screen


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.6-dev

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Latest Min-Vid add-on version (0.1.6-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Navigate to a website with videos supported by Min-Vid player (eg: YouTube, Vimeo, etc).
3. Send a video to the mini player and choose to play it.
4. Hover the Min-Vid player and observe the "Click here to move the thing" tooltip.

[Expected result]:
- The tooltip and bar appear on top of the video if the player is placed at the bottom of the screen and vice-versa.

[Actual result]:
- The tooltip and bar appear under the video shifting the player position upwards every time when hovered.

[Additional Notes]:
- As an idea, maybe we could restrict the dragging area only to the drag tooltip bar. In this way we can work around some issues that are caused by player dragging.
- Attached a screen recording of the issue http://goo.gl/D8aMoZ.

meandavejustice:
This will be changed into an icon which will be absolutely positioned over the video as a drag handle.

SoftVision-CiprianMuresan:
Good to know, because I've found another interesting issue regarding the drag bar http://goo.gl/b8OL9Z.

meandavejustice:
@SoftVision-CiprianMuresan yeah... that should go away once we get rid of that bar. yikes

meandavejustice:
@6a68 we can track the updates to the drag handle here

6a68:
@SoftVision-CiprianMuresan Could you file a separate bug for that second video?

> Good to know, because I've found another interesting issue regarding the drag bar http://goo.gl/b8OL9Z.

There are problems initiating drag on linux, but it so far hasn't done anything like that ^^ on windows or mac.

6a68:
@meandavejustice @johngruen OK, UX question: what if, instead of a drag handle, we show a wide border on hover & switch the pointer to the 'grab' hand to tell users the border is draggable, something like this:

![](https://www.dropbox.com/s/cvtuvl3hn5gyxey/Screen%20Shot%202016-09-06%20at%205.49.35%20PM.png?dl=0&raw=true)

[...]

PR #203, opened 2016-09-07
Make border draggable

6a68:
@meandavejustice Mind taking a look?

- Replaced drag handle with a draggable border area (technically padding) shown when the panel is hovered (including the invisible border area around the player window). Seems less janky, but still needs work.
- Replaced the "if not linux, make draggable" check with "if win or mac, make draggable", as discussed in #193.
- Refactored make-panel-draggable.js to remove the 'El' suffix on all the elements. I liked it at first, then decided it looked odd.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,8 +24,8 @@ const panel = require('sdk/panel').Panel({
   }
 });
 
-// check for linux here
-if (system.platform !== 'linux') {
+// Draggability seems to work for windows and mac, but not linux.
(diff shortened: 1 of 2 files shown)


meandavejustice:
@6a68 Would you mind attaching a screenshot or gif of this? I am unable to see it on my machine.

meandavejustice:
closing in favor of work being done on #195

Issue #184, opened 2016-09-03
min-vid doesn't play all the way to end of video?


pdehaan:
Using the XPI that you sent earlier in the week (0.1.1) and this random video (from https://reddit.com/r/videos):
https://www.youtube.com/watch?v=hU2jPw0mjaE

Play the video all the way to the end and you see: 
<img width="397" alt="_r_videos_and_downloads" src="https://cloud.githubusercontent.com/assets/557895/18221560/17f940f8-7134-11e6-8a3a-dd8ed8739af9.png">

Not sure why the video ends and only goes to 1:29 / 1:30, but I'll blame Maths + rounding.

---

Using a **min-vid@0.1.6** that I built locally, I still see this, although now the footer area fully expands horizontally, and the playback scrubber bar no longer has rounded corners:

<img width="360" alt="_r_videos_and_min-vid_doesn_t_play_all_the_way_to_end_of_video__ _issue__184_ _meandavejustice_min-vid" src="https://cloud.githubusercontent.com/assets/557895/18221613/3311d14c-7135-11e6-8d86-ee5cb2e9e6a8.png">

pdehaan:
Same w/ https://www.youtube.com/watch?v=Fxa3m6F89FQ
<img width="443" alt="_r_videos_and_min-vid_doesn_t_play_all_the_way_to_end_of_video__ _issue__184_ _meandavejustice_min-vid" src="https://cloud.githubusercontent.com/assets/557895/18221579/8aa60afa-7134-11e6-90cb-40885b435208.png">

meandavejustice:
@pdehaan apparently youtube embeds stop themselves right one second before completion. Thanks for reporting this, we also aren't canceling our stepping function when this happens. So we've been creating never ending intervals.

meandavejustice:
I've been messing with this on and off the past couple of days and youtube's iframe stuff just doesn't want to cooperate here. I'm going to fix the issue where the setInterval isn't being cancelled when youtube embeds end themselves, and put the actual "ends one second off" part off until all the other mvp issues are complete.

meandavejustice:
Moving this into `post-mvp` since the important part of this is being addressed in #207

pdehaan:
[...]

PR #207, opened 2016-09-08
Cancel requestAnimationFrame loop for ending youtube vids

meandavejustice:
- youtubes iframe player ends differently than normal videos,
it wants to end at one second before it's duration. This commit
addresses that. It is a hack but works.

- fixed issue where youtube videos were not showing replay/close
ui on video end.

- add back close method for replay/close ui view

- refs #184 (not a complete fix)

### DIFF
--- a/components/player-view.js
+++ b/components/player-view.js
@@ -4,6 +4,7 @@ const cn = require('classnames');
 
 const ytCtrl = require('../client-lib/yt-ctrl.js');
 const sendMetricsEvent = require('../client-lib/send-metrics-event.js');
+const sendToAddon = require('../client-lib/send-to-addon.js');
 const GeneralControls = require('./general-controls.js');
 
 function formatTime(seconds) {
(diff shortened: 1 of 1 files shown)


6a68:
looking now

components/player-view.js
```diff
@@ -26,9 +27,13 @@ module.exports = React.createClass({
       progress: currentTime / window.AppData.duration
     });
 
-    if (currentTime >= window.AppData.duration) {
-      window.AppData.playing = false;
-      window.AppData.playedCount++;
+    // TODO: the "-1" is a hack to force YouTube embeds to do what we want.
```
pdehaan: Is this the #184 issue? If so, we may want to add the bug number to the TODO comment for future context.

meandavejustice: good thinking @pdehaan will do.

6a68:
LGTM :+1:

Issue #224, opened 2016-09-12
Performance issues with the YouTube homepage opened


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.8-dev

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Latest Min-Vid add-on version (0.1.8-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Navigate to YouTube.
3. Browse the homepage for 15-20 mins, sending various videos to the Min-Vid player, load more videos on the YouTube homepage.
4. After this, scroll up and down the page, hover some videos, bring up context menu for videos and observe the browser behavior.

[Expected result]:
- Everything works as expected.

[Actual result]:
- The browser becomes sluggish, scrolling through the page shows a blank page and in some cases the whole browser content will become unresponsive.

[Additional Notes]:
- Tested also with uBlock Origin enabled and this causes the browser to become unresponsive faster.
- Attached a Gecko profile http://goo.gl/3oLun6.
- Attached a screen recording of the issue http://goo.gl/6UADlU.

SoftVision-CiprianMuresan:
On today's build (09.13.2016) the YouTube homepage is definitely better. Just to be sure, we should leave this issue opened for a few more day as we want to test with addons.

PR #225, opened 2016-09-12
Prevent overlay-icons from being inserted multiple times on interval

### LINKED ISSUE #223: The Min-Vid overlay gets progressively whiter on the YouTube homepage
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.8-dev

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:

[Steps to reproduce]:
2. Navigate to YouTube.
3. Wait for about a minute or two and observe the Min-Vid icon on the video thumbnails.

[Expected result]:
- The icon is gray and it is highlighted only on mouseover.

[Actual result]:
- The icon gets progressively whiter as time passes.

[Additional Notes]:
- Attached a screen recording (at 4x speed) of the issue http://goo.gl/qzJE2l.

meandavejustice:
- our checkForEmbeds methods in overlay-icon.js wasn't checking for
existance of the overlay icon before appending it. So we we're adding
multiple copies of the icon into the dom on an interval.

- fixes #223
- refs #224 surely this was affecting performance

### DIFF
--- a/data/icon-overlay.js
+++ b/data/icon-overlay.js
@@ -35,6 +35,8 @@ function ytEmbedChecks() {
 }
 
 function ytHomePageHandler(el) {
+  if (el.classList.contains('minvid__overlay__wrapper')) return;
+
   el.classList.add('minvid__overlay__wrapper');
   const tmp = getTemplate();
   tmp.addEventListener('click', function(ev) {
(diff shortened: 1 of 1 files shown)


6a68:
@meandavejustice It looks like this code keeps the interval running after the overlay icon has been inserted into the DOM. Is there any reason not to just `clearInterval` the first time the overlay icon is inserted?

meandavejustice:
@6a68 We run the interval so that it catches changes for both 1) when navigating around, since it's a single page app. 2) lazy loading on the home page

6a68:
Aha, makes sense. :+1: Merging

PR #264, opened 2016-09-16
Return volume to 50% on unmute

### LINKED ISSUE #260: Unmuting from speaker button wrongly keeps the sound off, If volume is previously reduced to 0 from slider
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.8-dev

[Affected Platforms]:
- All Windows

[Prerequisites]:

[Steps to reproduce]:
2. Navigate to YouTube.
3. Send a video to the Min-Vid player and play the video. (default volume is set at 50%)
4. Slide the Volume slider all the way to the left of the slider. 
5. Click "Unmute" button.

[Expected result]:
- The volume has returned to the last known value except 0, which is 50%.

[Actual result]:
- Volume button changes state from Muted to Normal but the volume is still at 0.

[Additional Notes]:
- The volume control should be like the one on YouTube.
- [Here](https://drive.google.com/file/d/0B6QogV2AJFKiUHl5X3hiZmRHQ1k/view?usp=sharing) is a video of the issue.

meandavejustice:
- fixes #260

### DIFF
--- a/components/player-view.js
+++ b/components/player-view.js
@@ -124,9 +124,18 @@ module.exports = React.createClass({
   unmute: function() {
     sendMetricsEvent('player_view', 'unmute');
 
+    let volumeEvt = {target: {value: '0.5'}};
+    // bump volume up to 50% if less than 5% since it is likely
+    // if the volume is < 5%, that the user dragged the volume
+    // slider down to 0% in order to mute.
+    if (parseInt(window.AppData.volume, 10) > 0.05) {
+      volumeEvt = {target: {value: window.AppData.volume}};
+    }
+
+    this.setVolume(volumeEvt);
+
     if (this.isYt) {
       ytCtrl.unmute();
-      ytCtrl.setVolume(window.AppData.volume * 100);
     } else {
       this.refs.video.muted = false;
     }


pdehaan:
If I had my volume at 100% previously, and hit Mute then Unmute, wouldn't this set the volume to 50% instead of my previous 100% level?

meandavejustice:
@pdehaan yes it would, I should make this a bit more robust before merging

pdehaan:
I'd expect if i was at a specific level before, unmuting would return me back to my previous volume.
But yeah, the obvious edge case there is if you did drag to 0% volume, then unmuting would be a no-op and we should probably default to 50% in that case.

meandavejustice:
@pdehaan What if we did something like:

```js
// window.AppData.volume here would be the past volume pre-muting
if (parseInt(window.AppData.volume, 10) > 0.25) {
  // set volume to old volume
} else {
  // set volume to 50%
}
```

We could also do lower than 25% above for the check

6a68:
@meandavejustice If you have the volume really low on purpose, then you mute momentarily, it would kinda suck to have the volume jump up to 50%.

What about just bumping the volume high enough that the slider blue is visible? Like, 5-10%, just so people aren't confused by clicking the mute button, but nothing happens?

6a68:
> If you have the volume really low on purpose, then you mute momentarily, it would kinda suck to have the volume jump up to 50%.

I guess that's not a valid use case: if the volume is non-zero, it'll be restored on unmute.

Meh, let's land it and see how it feels :beers:

PR #274, opened 2016-09-21
Position panel relative to screen by changing 'level' attribute

### LINKED ISSUE #253: Min-Vid Player position is affected by the resizing of the browser window
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.8-dev

[Affected Platforms]:
- All Windows

[Prerequisites]:

[Steps to reproduce]:
2. Resize the browser window to half the screen.
3. Drag and drop the Min-Vid Player outside the browser window to the left.
4. Horizontally resize the browser window towards the left.

[Expected result]:
- The browser window is resized and the Min-Vid player remains in its place.

[Actual result]:
- The Min-Vid player is moved towards the left together with the browser window.

[Additional Notes]:
- Attached a screen recording of the issue https://goo.gl/TP1LKC.

6a68:
@meandavejustice Hey, mind taking a look? This should be a quick one. I'm not sure you'll see any changes, since you are on an unusual window manager, but on Mac, this makes the experience much nicer. Commit message has more details:

----

This change ensures the panel doesn't move when the parent browser
window moves.

This change also sets the z-index so that the min-vid panel is shown
above all Firefox windows, and (on Mac) the panel is shown even if FF
loses application focus.

Fixes #253.

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,6 +36,7 @@ const panel = require('sdk/panel').Panel({
 panel.coords = { bottomOffset: -10, leftOffset: 10 };
 
 getActiveView(panel).setAttribute('noautohide', true);
+getActiveView(panel).setAttribute('level', 'top');
 
 // Draggability seems to work for windows and mac, but not linux.
 if (system.platform === 'winnt' || system.platform === 'darwin') {


6a68:
I was looking for something unrelated when I stumbled across this attribute definition on MDN: https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/panel.level

Pretty sweet one-liner bug fix ^_^

PR #317, opened 2016-09-27
Adds Test Pilot badge.

chuckharmston:


README.md
```diff
@@ -4,7 +4,7 @@
   </a>
 </p>
 
-[![Build Status](https://travis-ci.org/meandavejustice/min-vid.svg?branch=master)](https://travis-ci.org/meandavejustice/min-vid)
+[![Build Status](https://travis-ci.org/meandavejustice/min-vid.svg?branch=master)](https://travis-ci.org/meandavejustice/min-vid) [![Available on Test Pilot](https://img.shields.io/badge/available_on-Test_Pilot-0996F8.svg)](https://testpilot.firefox.com/experiments/)
```
pdehaan: Do we want to link to generic /experiments/ list page versus the [now fully public] experiment details page?

chuckharmston: It shouldn't be fully public until tomorrow. If so, that's a bug.

pdehaan: Fine, fine, @chuckharmston...
_TOMORROW_ when the experiments pages are public, should the badge point to a generic /experiments/ list, or to the details page?

pdehaan: Also, something funny is going on for me.
In both Chrome and Firefox, if I click on the badge, it goes to https://testpilot.firefox.com/experiments/ and then quickly redirects to https://testpilot.firefox.com/. Clicking "Back" in my browser takes me to /experiments/ which again quickly redirects me back to the main Test Pilot page. I have to hold down the back button so I can jump back to GitHub to avoid the redirect loop.

Not sure why it isn't letting me look at /experiments/ if I don't have the Test Pilot add-on installed.
But I'm almost certain that I shouldn't be bitching about it in a nested comment in a closed PR in a tangential repo, so perhaps I'll file a new bug in **mozilla/testpilot**.

pdehaan: Actually, mildly interesting and now slightly more related to this repo:

- https://testpilot.firefox.com/experiments works.
- https://testpilot.firefox.com/experiments/ redirects to https://testpilot.firefox.com/.

PR #318, opened 2016-09-27
Drop played_count metric

### LINKED ISSUE #313: Let's drop `played_count`
`played_count` is in the documentation: https://github.com/meandavejustice/min-vid/blob/master/docs/metrics.md

It's a per-video count of the number of times a video reaches the end.  We couldn't think of a reason to have that so we're suggesting we drop it.

meandavejustice:
- fixes #313


6a68:
@meandavejustice good to merge after a rebase

PR #365, opened 2016-09-29
Remove unneeded React bindings and fix ESLint error

pdehaan:


components/app-view.js
```diff
@@ -7,12 +8,12 @@ const ErrorView = require('./error-view');
...
   render: function() {
     return (
-        <div className={'app'}>
+        <div className='app'>
           {/* Show Error View, ELSE Show Loading View ELSE no view */}
           {this.props.error ? <ErrorView {...this.props}/> :
             (!this.props.loaded) ? <LoadingView {...this.props}/> : null}
 
-          <div className={this.props.loaded ? 'player-wrap' : 'player-wrap hidden'}>
+          <div className={cn('player-wrap', {hidden: !this.props.loaded})}>
```
pdehaan: Switched to `classnames` module usage here since we have it elsewhere.

components/error-view.js
```diff
@@ -30,17 +30,18 @@ module.exports = React.createClass({
...
             <GeneralControls {...this.props} />
           </div>
           <div className='error-message-container'>
-            <p className='error-message'>{"Something's gone wrong with this video, Try again later."}
+            <p className='error-message'>
+              Something's gone wrong with this video, Try again later.
               <br/>
               <br/>
-              <span className='error-link' onClick={this.sendToTab}>{"Open in new tab"}</span>
+              <span className='error-link' onClick={this.sendToTab}>Open in new tab</span>
```
pdehaan: This was actually the line that was throwing some ESLint chaos for me since it was using double quotes in the React binding. And why I went down the rabbit hole of removing the other string bindings.

meandavejustice:
Thank you @pdehaan !

PR #398, opened 2016-10-06
Close fullscreen youtube player when video is opened in min-vid

### LINKED ISSUE #367: fullscreen youtube should go to a normal size when minvid is turned on
it looks a bit weird when minvid is top of fullscreen youtube video, also it would be nice for navigation as you dont' have to click extra button to exit the fullscreen

danrr:
- fixes #367

### DIFF
--- a/data/icon-overlay.js
+++ b/data/icon-overlay.js
@@ -90,6 +90,7 @@ function ytWatchElementHandler(el) {
     evNoop(ev);
     const videoEl = document.querySelector('video');
     videoEl.pause();
+    closeFullscreen();
     self.port.emit('launch', {
       url: window.location.href,
       domain: 'youtube.com',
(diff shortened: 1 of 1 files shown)


data/icon-overlay.js
```diff
@@ -187,3 +188,9 @@ function evNoop(ev) {
   ev.preventDefault();
   ev.stopImmediatePropagation();
 }
+
+function closeFullscreen() {
+  if(document.mozFullScreenEnabled){
```
meandavejustice: Please update this line to follow code conventions throughout the rest of the project.
* space after `if` statement
* space before curly bracket
```js
if (document.mozFullScreenEnabled) {
```
We should add this to our `eslintrc` so it doesn't show down our review process.

danrr: Cool, it's sorted now.

PR #404, opened 2016-10-07
Pass video volume and muted state to video player to keep audio settings

### LINKED ISSUE #349: Popped out video audio levels should match original video.
I muted a video and popped it out and the audio immediately started playing.
Also tested this with just a modified audio level and it also was not reflected in the windowed video.

We should work to reflect the audio level at the time the user clicks the button to pop it out.

danrr:
- fixed #349

### DIFF
--- a/data/icon-overlay.js
+++ b/data/icon-overlay.js
@@ -94,7 +94,9 @@ function ytWatchElementHandler(el) {
     self.port.emit('launch', {
       url: window.location.href,
       domain: 'youtube.com',
-      time: videoEl.currentTime
+      time: videoEl.currentTime,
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
meandavejustice: This seems like it should be in your global `.gitignore`

danrr:
Yeah, you're right. I've taken out the commit.

.gitignore
danrr: Yeah, you're right. I've taken out the commit.

PR #435, opened 2016-10-17
Increase minimized height on Linux

### LINKED ISSUE #427: Minimized height in linux
The height of the player when minimized appears to be shorter than the buttons.
screenshot at 2016-10-14 20-35-29
If I replace `minimizedHeight: 40` with `minimizedHeight: 50`, it seems to look better.
screenshot at 2016-10-14 20-34-35
These screenshots seem to be consistent with those found in issue #286. I think it has to do with the small triangle found in the top left corner of the window, which I presume is included in the dimensions.
The current machine im using has Linux Mint 18 installed on it, but it appears to be the same across all linux boxes.

vickychijwani:
This should fix #427 (tested on Debian sid).

Let me know if there's a better way to fix this that doesn't involve a platform-specific check or a hard-coded 10 pixel increment (maybe there's a constant somewhere that stores the height of the small triangle on the top-left?).

### DIFF
--- a/lib/panel-utils.js
+++ b/lib/panel-utils.js
@@ -79,6 +79,9 @@ const DEFAULT_DIMENSIONS = {
   width: 320,
   minimizedHeight: 40
 };
+if (system.platform === 'linux') {
+  DEFAULT_DIMENSIONS.minimizedHeight += 10;
+}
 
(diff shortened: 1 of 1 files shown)


meandavejustice:
👍 Thanks @vickychijwani

PR #436, opened 2016-10-17
Make 'send to tab' send to the original tab, if open

### LINKED ISSUE #369: 'send to tab' should send the video to same tab it was open from

vickychijwani:
This fixes #369.

### DIFF
--- a/lib/panel-utils.js
+++ b/lib/panel-utils.js
@@ -8,6 +8,7 @@
 
 const system = require('sdk/system');
 const events = require('sdk/system/events');
+const getVideoId = require('get-video-id');
 const { getActiveView } = require('sdk/view/core');
 const sendMetricsData = require('./send-metrics-data.js');
 const makePanelDraggable = require('./make-panel-draggable.js');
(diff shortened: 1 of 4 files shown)


lib/panel-utils.js
```diff
@@ -324,6 +317,32 @@ function messageHandler(opts) {
...
+    const tabs = require('sdk/tabs');
+    let targetTab = null;
+    for (let tab of tabs) { // eslint-disable-line prefer-const
+      if (tab.id === opts.tabId) {
+        targetTab = tab;
+        break;
+      }
+    }
+    if (targetTab !== null) {
+      targetTab.url = pageUrl;
```
meandavejustice: this should be `targetTab.activate()`
the above code only refreshes the tab, it doesn't make it active.

meandavejustice:
It would be nice if this also supported the case where a video is launched from an embed or context menu. And then opened from that link.

In other words, it would be good if we could check the actual urls to see if we have the page open, instead of solely storing a reference to the tab.

vickychijwani:
@meandavejustice handled that case now by checking that the video id matches - since the URLs could have slightly different parameters like time.

lib/panel-utils.js
```diff
@@ -324,6 +318,36 @@ function messageHandler(opts) {
...
+      if (tab.id === opts.tabId) {
+        targetTab = tab;
+        break;
+      }
+    }
+
+    const targetVideoId = getVideoId(targetTab.url);
+    const pageVideoId = getVideoId(pageUrl);
+    if (targetTab !== null && pageVideoId && targetVideoId === pageVideoId) {
+      targetTab.url = pageUrl;
```
vickychijwani: I'm not sure what the intended behaviour is here. Do we want to simply close the min-vid panel and activate the tab, leaving it up to the user to seek to the desired time and play it? Or do we want to auto-play the video from the exact time the panel was closed? I was assuming the latter, which is why I'm setting the URL here.

meandavejustice: ahh I see, the original intention was to simple activate the already opened tab. You make a good point about continuing playback though. Somehow I wasn't taking that into account. I'll think about it a bit and play with it over the weekend and see what feels best. Thanks!

meandavejustice: after thinking about it a bit, i think this is the best solution. Thank you @vickychijwani

vickychijwani:
Rebased and resolved the merge conflicts.

Issue #113, opened 2016-07-29
Warning: setState(...): Can only update a mounted or mounting component


pdehaan:
Looking at https://vimeo.com/176291464 and after sending a video to the Min-Vid player and minimizing the player, I somehow got this message in my Browser Console (Tools > Web Developer > Browser Console)

> Warning: `setState(...)`: Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
> 
> &mdash; via bundle.js:20922

![jar_file____users_pdehaan_library_application_20support_firefox_profiles_118pl5yo_activity-stream-recommendations_extensions__min-vid_xpi__data_bundle_js](https://cloud.githubusercontent.com/assets/557895/17234095/5bd6d576-54ea-11e6-9be6-428f9ec8b1fa.png)

meandavejustice:
Looks like an upstream bug. I commented on it here, https://github.com/wwayne/react-tooltip/issues/148#issuecomment-235926176. Just forgot to link it in my pr

rugk:
I can reproduce this too.

```
console.error: min-vid: 
  Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
console.error: min-vid: 
  Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
console.error: min-vid: 
  Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component.
```

meandavejustice:
No longer seeing this, closing

PR #458, opened 2016-10-21
add L10N support

### LINKED ISSUE #110: Only the "Play" and "Close" buttons have tooltips in the minimized version.
from @pdehaan in https://github.com/meandavejustice/min-vid/pull/94#issuecomment-235444070

youtube
Only the "Play" and "Close" buttons have tooltips in the minimized version.

### LINKED ISSUE #441: Add L10N support
Current strings:
* Tooltips(throughout)
* [loading view](https://github.com/meandavejustice/min-vid/blob/master/components/loading-view.js#L27)
* [error view](https://github.com/meandavejustice/min-vid/blob/master/components/error-view.js#L42)

[sdk/l10n](https://developer.mozilla.org/en-US/Add-ons/SDK/High-Level_APIs/l10n)

meandavejustice:
- fixes #441
- fixes #110
- refs #113 got rid of some warnings, still being caused in
general-controls component
- add `locales/en-US.properties`
- add `lib/locale-strings.js` for mapping keys to strings
- add `client-lib/merge-deep.js` as a deep clone replacement for
`Object.assign`
- update all strings to use localized strings
- switched to using "data-for" syntax with `react-tooltip`. This
allows for us to pass in the localized strings without issues
and also silences some of the warnings mentioned in #113
- removed excess `<style>` tag in data/default.html

#### note
had to add `unsafe-content-script` permission in order to
 store a "strings" object on unsafeWindow.AppData. I think this is
okay for now, it's not causing any issues with e10s on my machine.
I'm planning to rework the way we are handling data updates one
day, so it should be fine for now.

### DIFF
--- a/locale/en-US.properties
+++ b/locale/en-US.properties
@@ -1,0 +1,12 @@
+tooltip_play= Play
+tooltip_pause= Pause
+tooltip_mute= Mute
+tooltip_unmute= Unmute
+tooltip_send_to_tab= Send to tab
+tooltip_minimize= Minimize
+tooltip_maximize= Maximize
+tooltip_close= Close
+error_msg= Something's gone wrong with this video. Try again later.
+error_link= Open in new tab
+# LOCALIZATION NOTE: placeholder is domain name
+loading_msg= Loading video from %s
--- a/.jpmignore
+++ b/.jpmignore
@@ -3,6 +3,7 @@
 !/data/**
 !/docs/images/dark-logo.png
 !/lib/**
+!/locale/**
 !/index.js
 !/install.rdf
 !/bootstrap.js
--- a/app.js
+++ b/app.js
@@ -14,7 +14,8 @@ const defaultData = {
   duration: 0,
   progress: 0.001, // force progress element to start out empty
   playing: false,
-  volume: '0.5'
+  volume: '0.5',
+  strings: {}
 };
 
 window.AppData = new Proxy(defaultData, {
--- a/components/error-view.js
+++ b/components/error-view.js
@@ -1,6 +1,5 @@
 const React = require('react');
 const cn = require('classnames');
-const ReactTooltip = require('react-tooltip');
 const sendToAddon = require('../client-lib/send-to-addon');
 const sendMetricsEvent = require('../client-lib/send-metrics-event');
 const GeneralControls = require('./general-controls');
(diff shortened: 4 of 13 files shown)


meandavejustice:
@pdehaan @6a68 r?

locale/en-US.properties
```diff
@@ -0,0 +1,11 @@
...
+tooltip_pause= Pause
+tooltip_mute= Mute
+tooltip_unmute= Unmute
+tooltip_send_to_tab= Send to tab
+tooltip_minimize= Minimize
+tooltip_maximize= Maximise
+tooltip_close= Close
+error_msg= Something's gone wrong with this video, Try again later.
+error_link= Open in new tab
+loading_msg= Loading video from
```
6a68: I'm not sure this will work. This assumes that the domain would come after the phrase 'loading video from', but in right-to-left languages, this won't be the case.

In cases like this, you usually want a placeholder, like `${domain}`, so that localizers can figure out what the complete phrase means, then move the variable around accordingly. You can also insert a comment just above this particular line to explain to localizers what the variable names mean. To work with the localization tools, comments should start with `# LOCALIZATION NOTE:`, like [this](https://github.com/mozilla/testpilot/blob/ab37969de5f7eb52a222f37fee283cf6e773da97/locales/en-US/app.ftl#L59) example from test pilot.

Since we have no locales yet, I'm good to land this for now and file a followup bug to localize it correctly.

locale/en-US.properties
```diff
@@ -0,0 +1,11 @@
+tooltip_play= Play
+tooltip_pause= Pause
+tooltip_mute= Mute
+tooltip_unmute= Unmute
+tooltip_send_to_tab= Send to tab
+tooltip_minimize= Minimize
+tooltip_maximize= Maximise
+tooltip_close= Close
+error_msg= Something's gone wrong with this video, Try again later.
```
6a68: Either replace `video,` with `video.` or replace `Try` with `try`

lib/locale-strings.js
```diff
@@ -0,0 +1,15 @@
+const _ = require('sdk/l10n').get;
+
+module.exports = {
+  errorMsg: _('error_msg'),
+  errorLink: _('error_link'),
+  loadingMsg: _('loading_msg'),
```
6a68: You could use interpolation here to insert a variable that you'd then replace with the domain name in the React code, like: `loadingMsg: _('loading_msg', '$domain')`. Then the React code could just do a simple string replace on '$domain'.

client-lib/merge-deep.js
```diff
@@ -0,0 +1,19 @@
...
+      if (isObject(source[key])) {
+        if (!target[key]) Object.assign(target, { [key]: {} });
+        mergeDeep(target[key], source[key]);
+      } else {
+        Object.assign(target, { [key]: source[key] });
+      }
+    }
+  }
+  return target;
+}
```
meandavejustice: There definitely is, I'll switch it over.

lib/locale-strings.js
meandavejustice: Good point

Issue #194, opened 2016-09-06
Volume setting is not remembered for next videos even if the button state is


SoftVision-CiprianMuresan:
[Affected versions]:
- Firefox 45.0 and up
- Min-Vid 0.1.6-dev

[Affected Platforms]:
- All Windows
- All Mac 
- All Linux

[Prerequisites]:
- Latest Min-Vid add-on version (0.1.6-dev) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites. 
2. Navigate to a website with videos supported by Min-Vid player (eg: YouTube, Vimeo, etc).
3. Send a video to the mini player and choose to play it.
4. Mute the video.
5. Send a different (or the same) video to the player and choose to play it.
6. Observe the volume setting icon.

[Expected result]:
- The volume setting and icon are are remembered / reset to their defaults (depending on what we want)

[Actual result]:
- The volume setting icon is in its muted state, but the audio is not muted.

[Additional Notes]:
- Depending on how we want the player to perform, we can either have it remember volume settings or reset them every time the player is closed and reopened.
- The issue is also reproducible if the player is restarted.
- The YouTube and Vimeo players remember volume settings which you set on a video even if the browser is restarted.
- Attached a screen recording of the issue http://goo.gl/hhTfng.

SoftVision-CiprianMuresan:
The issue is no longer reproducible on the latest build (0.1.9).

PR #475, opened 2016-11-01
Draggable, always on top window

6a68:
@meandavejustice Please give this a spin and see what you think.

The major problem with this patch is that clicks aren't passed through to the iframe on linux. This makes it impossible to dismiss ads shown over top of the video.

Two possible alternatives would be (1), bring back the drag handle, or (2), only allow dragging from elements that have a 'drag' class. We could set this 'drag' class on the controls bar, but not on the individual controls, so that you'd drag the shaded background of the controls bar, but could still click to hide ads.

Thoughts / suggestions welcome :-)


6a68:
I'm going to rebase and try my second idea from the description (making the shaded controls area draggable).

6a68:
@Softvision-VladBacia Hi, you mentioned you were getting all Firefox windows always-on-top using this branch and nightly. What OS/distro, version, window manager are you using?

6a68:
@Softvision-VladBacia I'm not able to repro your always-on-top bug using current nightly in 16.04 ubuntu (unity) or ubuntu gnome 16.04. Still need to test kubuntu (ubuntu + kde).

6a68:
@meandavejustice OK, I've updated the dragging code. Now the controls area is draggable, but not the controls themselves. This allows ads shown inside the iframe to be dismissed, but still provides a nice dragging experience.

Comments welcome :-)

6a68:
Note: this branch causes a crash in Nightly + Linux (ubuntu 16.04 x64) when the player window is closed. I think it's a symptom of a known vmware/FF bug (https://bugzil.la/1298692), will try on hardware tomorrow.

6a68:
Wil discovered that the minvid window is actually resizable on linux (though not mac, haven't tried win). So that's an interesting easter egg ^_^

Softvision-VladBacia:
@6a68 , I was using Windows 8.1 x64 with Nightly x64 (Build ID: 20161101030207).

meandavejustice:
@6a68 Would you mind adding a `cursor: move;` for the controls bar

Softvision-VladBacia:
Also, while dragging there are a lot of errors like:
```
Error: Permission denied to access property "target" 1 bundle.js:20138:7
	getEventTarget resource://min-vid/data/bundle.js:20138:7
	handleTopLevelImpl resource://min-vid/data/bundle.js:14067:27
	[155]</</Mixin.perform resource://min-vid/data/bundle.js:19363:13
	[106]</ReactDefaultBatchingStrategy.batchedUpdates resource://min-vid/data/bundle.js:13174:7
	batchedUpdates resource://min-vid/data/bundle.js:17069:3
	[113]</ReactEventListener.dispatchEvent resource://min-vid/data/bundle.js:14159:7
	bound  self-hosted
```
and
```
Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the undefined component. 1 bundle.js:1904:9
	printWarning resource://min-vid/data/bundle.js:1904:9
	warning resource://min-vid/data/bundle.js:1928:9
	getInternalInstanceReadyForUpdate resource://min-vid/data/bundle.js:16798:47
	[136]</</ReactUpdateQueue.enqueueSetState resource://min-vid/data/bundle.js:16945:28
	[79]</</ReactComponent.prototype.setState resource://min-vid/data/bundle.js:8630:3
	showTooltip resource://min-vid/data/bundle.js:2662:7
	bindListener/< resource://min-vid/data/bundle.js:2586:9
	bindListener resource://min-vid/data/bundle.js:2574:7
	componentDidMount resource://min-vid/data/bundle.js:2519:7
	invokeComponentDidMountWithTimer resource://min-vid/data/bundle.js:8992:3
	[53]</</<.notifyAll resource://min-vid/data/bundle.js:4058:9
	[132]</ON_DOM_READY_QUEUEING.close resource://min-vid/data/bundle.js:16341:5
	[155]</</Mixin.closeAll resource://min-vid/data/bundle.js:19429:11
	[155]</</Mixin.perform resource://min-vid/data/bundle.js:19376:11
	batchedMountComponentIntoNode resource://min-vid/data/bundle.js:14631:3
	[155]</</Mixin.perform resource://min-vid/data/bundle.js:19363:13
	[106]</ReactDefaultBatchingStrategy.batchedUpdates resource://min-vid/data/bundle.js:13174:7
	batchedUpdates resource://min-vid/data/bundle.js:17069:3
	[121]</</ReactMount._renderNewRootComponent resource://min-vid/data/bundle.js:14794:5
	[121]</</ReactMount._renderSubtreeIntoContainer resource://min-vid/data/bundle.js:14873:21
	[121]</</ReactMount.render resource://min-vid/data/bundle.js:14894:12
	renderApp resource://min-vid/data/bundle.js:37:3
	[1]</window.AppData<.set resource://min-vid/data/bundle.js:24:5
```

6a68:
@meandavejustice I could bring back the drag handle instead of dragging the controls bar, that was just one random interaction idea. A drag handle would probably make more sense, since the other controls have icons.

meandavejustice:
@6a68 I think I like dragging the control bar better. It's kind of like a title bar so it feels pretty natural. I never see drag icons in the way we had. We just need a way to let the user know that it is draggable, so the cursor should suffice.

lib/vendor/noitidart-ostypes/cutils.jsm
```diff
@@ -0,0 +1,462 @@
+/*** CTypes Utilities ***
+ * Stuff that I find useful in my CTypes work under all Operating Systems
```
pdehaan: Any benefits from copy-pasta-ing the [~12,400 lines of vendored code] into our repo versus adding https://github.com/Noitidart/ostypes as a [dev]dependency and copying the node_modules files into ./lib/vendor/*?

pdehaan: Or seeing if we can get the ostypes module added to npm maybe.

pdehaan: Alright, ignore me. There isn't a package.json file in that upstream repo, so npm chokes pretty hard. I think before we were resorting to using bower to install some of these non-npm GitHub modules.

FWIW, I filed https://github.com/Noitidart/ostypes/issues/9. We could possibly even craft up a bare bones package.json and submit a PR and see if that is easier to get merged quickly, if we do want to go the npm dependency route.

meandavejustice: I'd like to see this as a dev-dep installed with npm. If they don't wanna accept a package.json, then I'd rather just have this checked in. Pretty much anything to avoid using two different package managers.

Noitidart:
> The major problem with this patch is that clicks aren't passed through to the iframe on linux. This makes it impossible to dismiss ads shown over top of the video.

> Two possible alternatives would be (1), bring back the drag handle, or (2), only allow dragging from elements that have a 'drag' class. We could set this 'drag' class on the controls bar, but not on the individual controls, so that you'd drag the shaded background of the controls bar, but could still click to hide ads.

Can we use a MouseEvent listener and check `e.target` has a certain class, if it doesnt then we let the click through?

lib/vendor/noitidart-ostypes/cutils.jsm
pdehaan: PR created and already merged! ❤️ https://github.com/Noitidart/ostypes/pull/10

We should be able to do this now with `npm i Noitidart/ostypes -D` (until it gets published on npm, at which point we can just do `npm i ostypes -D`). But we'd still need an npm postinstall hook to copy the .jsm files from ./node_modules/ostypes/\*.jsm to ./lib/vendor/noitidart-ostypes/\*.

6a68:
> Can we use a MouseEvent listener and check e.target has a certain class, if it doesnt then we let the click through?

@Noitidart Great idea. The problem is that we don't know the class names associated with any random website's display ads ahead of time, so that approach is a bit brittle. The other issue is that we'd definitely be violating terms of service if users couldn't click on ads shown on top of a video ;-)

I think the draggable controls bar works pretty well. I've pushed it to a different branch (window-player-updated) while I work on rebasing against the audio stuff that landed in master. Give it a try, comments welcome: https://github.com/6a68/min-vid/tree/window-player-updated

Noitidart:
Thanks @6a68 that's true, but is it possible to do the reverse? Say on all our elements with class .min-vid-el and then if mouseevent doesnt detect the e.target has min-vid-el then let the click through?

6a68:
@Noitidart Yup! That's exactly what we do now: look for els where `e.target.classList.contains('drag')`, and that class is applied to the controls bar.

In the first version of this branch, I tried making the whole window draggable, so there was no easy way to differentiate between a mousedown intended to dismiss an ad, and a mousedown intended to start dragging the window. We probably could have compared the coords of the mousedown event to the current location of the mouse pointer, or something, to differentiate drags from clicks--but that would likely have required more ctypes code, which I wanted to avoid.

BTW, your ctypes code saved me weeks of struggle. Would you be interested in being added to the list of contributors on the left side of the [min vid experiment page](https://testpilot.firefox.com/experiments/min-vid)?

lib/vendor/noitidart-ostypes/cutils.jsm
6a68: > We should be able to do this now with npm i Noitidart/ostypes -D (until it gets published on npm, at which point we can just do npm i ostypes -D). But we'd still need an npm postinstall hook to copy the .jsm files from ./node_modules/ostypes/*.jsm to ./lib/vendor/noitidart-ostypes/*.

@pdehaan That's a great idea! Would you be interested in opening a PR against this branch?

Noitidart:
Thanks @6a68 that would be so so cool to be on the contributors column!

If you would like to make the whole window draggable (except when on controls/etc) I would love to give it a shot. Let me know please if you want to go that way.

Softvision-VladBacia:
#369 is not included in https://github.com/6a68/min-vid/tree/window-player-updated . Is this expected?

Softvision-VladBacia:
With the latest min-vid xpi from window-player-updated branch, I see the below error in the browser console when sending a video to min-vid player.

```
min-vid: TypeError: can't pass undefined to argument 2 of bool SetWindowPos(void*, void*, int, int, int, int, unsigned_int)
Stack trace:
_winntTopify@resource://gre/modules/commonjs/toolkit/loader.js -> resource://min-vid/lib/topify.js:59:21
topify@resource://gre/modules/commonjs/toolkit/loader.js -> resource://min-vid/lib/topify.js:50:41
create/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://min-vid/lib/window-utils.js:62:21
whenReady@resource://gre/modules/commonjs/toolkit/loader.js -> resource://min-vid/lib/window-utils.js:29:60
whenReady/<@resource://gre/modules/commonjs/toolkit/loader.js -> resource://min-vid/lib/window-utils.js:30:22
notify@resource://gre/modules/commonjs/toolkit/loader.js -> resource://gre/modules/commonjs/sdk/timers.js:40:9
 1 timers.js:43
```

Noitidart:
@Softvision-VladBacia this error is due to use of `ostypes.CONST.HWND_TOP`, after this commit is reverted this error will go away.

6a68:
Rebased and works (draggable, topmost on windows, mac, linux), except for these issues:

- [ ] l10n strings are missing
- [ ] 1px white border on windows
- [ ] CSS seems broken (domain in player view looks like a serif font)
- [ ] the 'return to tab' patch needs to be rewritten: old code was in panel-utils (which is gone)
- [ ] tooltips all say 'true', possibly only because l10n is missing

Softvision-VladBacia:
@6a68 , I was taking a look an xpi built from `window-player-squashed` branch and I have noticed that issue #194 is reproducible. 

My question is, should we test the branches related to Min-Vid from your git, or should we focus on master branch and localization and wait for a request when we need to test from your git?

6a68:
Turns out windows remember their screen position by default, so that's a bit more code we don't have to write.

@meandavejustice Hey, I think this may be ready to land. Rebased against today's master and re-squashed. Mind taking a look?

meandavejustice:
Still seeing this error
```
Error: Permission denied to access property "target"
```
The stack trace isn't helpful as it's all in react code..
```
getEventTarget resource://min-vid/data/bundle.js:19954:7
	handleTopLevelImpl resource://min-vid/data/bundle.js:13921:27
	[156]</Mixin.perform resource://min-vid/data/bundle.js:19190:13
	[107]</ReactDefaultBatchingStrategy.batchedUpdates resource://min-vid/data/bundle.js:13034:7
	batchedUpdates resource://min-vid/data/bundle.js:16902:3
	[114]</ReactEventListener.dispatchEvent resource://min-vid/data/bundle.js:14013:7
	bound  self-hosted
```
 
May not be too big of a deal, but I'd like to figure out what is happening so it doesn't bite us after pushing the update.

6a68:
One bug remaining: need to detect when the window is closed via hotkeys, to avoid the 'dead object' error.

6a68:
@meandavejustice OK, I've got the dead window errors handled, I think. Mind giving it another try?

In case you're curious, no DOM event actually worked. I had to resort to listening for _any_ XUL window to be destroyed, then look through the list of remaining windows for the minvid window: https://github.com/6a68/min-vid/commit/ad8e92b

PR #482, opened 2016-11-07
Add support for pontoon and script for processing l10n files

meandavejustice:
- add `glob@7.1.1` dependency
- add `bin/locales` script for handling l10n files
- add locales/en-US/addon.properties for pontoon


meandavejustice:
cc/ @clouserw

6a68:
looking now

bin/locales
```diff
@@ -0,0 +1,23 @@
...
+
+glob('locales/*/*.properties', function (err, files) {
+  if (err) return console.error('Error finding files', err);
+  files.forEach(function(path) {
+    fs.createReadStream(path).pipe(fs.createWriteStream(`locale/${parseLang(path)}.properties`));
+  });
+});
+
+function parseLang(path) {
+  const segments = path.split('/');
```
pdehaan: Not that we probably care right now, but I don't think this will work on Windows.

We could potentially use the `path` module and use a combo of `path.dirname()` and `path.sep`, a la:

```js
function parseLang(file) {
  return path.dirname(file)
    .split(path.sep)
    .pop();
}

**Comment by pdehaan** [comment]
For my own curiosity, what's the reason we're putting the files in _./locales/en-US/addon.properties_ and then copying the files into _./locale/en-US.properties_? Why not just put them in the target directory directly and not need the intermediate build step?

I was initially thinking we were merging locale .properties files, but it turns out if I have the following that the .properties files get messed up:

**./locales/en-US/addon.properties**
```ini
tooltip_play= Play
tooltip_pause= Pause
tooltip_mute= Mute
tooltip_unmute= Unmute
tooltip_send_to_tab= Send to tab
tooltip_minimize= Minimize
tooltip_maximize= Maximize
tooltip_close= Close
error_msg= Something's gone wrong with this video. Try again later.
error_link= Open in new tab
# LOCALIZATION NOTE: placeholder is domain name
loading_msg= Loading video from %s
```

**./locales/en-US/site.properties**
```ini
name= peter
```

Then if I run `npm run locales`, I get the following **./locale/en-US.properties**:

```ini
name= petery= Play
tooltip_pause= Pause
tooltip_mute= Mute
tooltip_unmute= Unmute
tooltip_send_to_tab= Send to tab
tooltip_minimize= Minimize
tooltip_maximize= Maximize
tooltip_close= Close
error_msg= Something's gone wrong with this video. Try again later.
error_link= Open in new tab
# LOCALIZATION NOTE: placeholder is domain name
loading_msg= Loading video from %s
```

6a68:
@meandavejustice :+1: lgtm, works, just going to merge it and file a followup bug for windows support in the locale script (great catch, @pdehaan).

bin/locales
meandavejustice: Jared filed https://github.com/meandavejustice/min-vid/issues/483 for this

Issue #483, opened 2016-11-08
Make bin/locales Windows-compatible


6a68:
From @pdehaan's comment in https://github.com/meandavejustice/min-vid/pull/482#discussion_r86875494:

`path.split('/')` won't work with Windows. The path separator needs to be abstracted out.

pdehaan:
TL;DR, we should probably use the OS-aware [`path.sep`](https://nodejs.org/api/path.html#path_path_sep) (and then tweak the code to not use `path` as a variable name).
Ref: [./bin/locales#L21](https://github.com/meandavejustice/min-vid/blob/15da7648ba2f3527deb40bb4478e7849712154f2/bin/locales#L21).

pdehaan:
Actually, this also seems to fail hard on Windows on `./bin/locales` command. I think that may need to be converted to `"node ./bin/locales"`:

```sh
C:\Users\pdehaan\dev\github\min-vid>node --version
v6.9.1

C:\Users\pdehaan\dev\github\min-vid>npm run package

> min-vid@0.3.3 prepackage C:\Users\pdehaan\dev\github\min-vid
> npm run locales && npm run lint && svgo data/img


> min-vid@0.3.3 locales C:\Users\pdehaan\dev\github\min-vid
> ./bin/locales

'.' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.10586
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "locales"
npm ERR! node v6.9.1
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! min-vid@0.3.3 locales: `./bin/locales`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the min-vid@0.3.3 locales script './bin/locales'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the min-vid package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     ./bin/locales
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs min-vid
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls min-vid
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
[...]

PR #486, opened 2016-11-10
Add cross-env and fix some Windows issues w scripts

pdehaan:
Doesn't fix the `path.sep` issue, but I'm not seeing that on my Windows 10 cmd.exe config.

Ref #483

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
   },
   "scripts": {
     "lint": "eslint . bin/*",
-    "build-script": "NODE_ENV=production browserify app.js -t [ babelify --presets [ react ] ] | uglifyjs -c > data/bundle.js",
-    "watch-script": "NODE_ENV=development watchify app.js -o data/bundle.js -t [ babelify --presets [ react ] ]",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,8 +32,8 @@
     "prepackage": "npm run locales && npm run lint && svgo data/img",
     "prepush": "npm run package",
     "dev": "npm run locales && npm run watch-script & npm run watch & http-server -c-1",
```
6a68: does `http-server ` need a `node` in front of it?

6a68: oh, of course not - it's still on the path, just a windows path.

6a68:
Thanks @pdehaan! Merging :tada:

package.json
pdehaan: As far as I know, `http-server` is a package, not a ./bin/ script, see [package.json:54](https://github.com/meandavejustice/min-vid/blob/327820920aeb1e387e383c8372fe3fd2c62e2157/package.json#L54).

PR #498, opened 2016-11-16
497 draggable views

6a68:
@meandavejustice Mind taking a look? makes the error and loading views draggable, and fixes the annoying selectable text in the error view


6a68:
eh, just gonna land it :tada:

PR #512, opened 2016-11-23
add build-addon.sh

clouserw:


bin/build-addon.sh
```diff
@@ -0,0 +1 @@
+npm run deploy
```
6a68: @clouserw I think you want `npm install && npm run deploy`

6a68:
@clouserw Based on the CI error, looks like you'll also need to add `bin/build-addon.sh` to `.eslintignore` at the top of the repo

6a68:
lgtm :+1:

PR #519, opened 2016-12-07
Update min-vid URL to use the new automatically built packages

clouserw:
When we're ready to start using the automatic builds, let's land this (and the accompanying patch on t.f.c)


clouserw:
requesting review from meandavejustice since he's controlling the other end and I'm out next week.  If you don't want to do this until I get back, that's cool too, we can close this and reopen it then or whatever.

PR #524, opened 2016-12-13
<audio>, <video>, and soundcloud support

### LINKED ISSUE #442: Basic <video> support
We want to add support for `<video>` tags by pulling the direct url from the `src` attribute.

We will probably need to write a regular expression to check that the url is formatted correctly and not using the `mediasource://` protocol.

### LINKED ISSUE #443: Add support for <audio> tags
So this issue includes:
- adding context-menu support for `<audio>` tags
- adding context-menu support for direct links to audio files
- update the player to handle playback of audio

This issue only includes the most basic launching/playback. Further ui issues will be addressed in other issues.

meandavejustice:
### changed
* add support for <audio> tags with direct links to audio files (via context menu)
* add support for <video> tags with direct links to video files (via context menu)
* add support for soundcloud tracks (via context menu)
* convert all components to es6 classes
* refactor to using a global event emitter for handling state changes
* pull all logic for individual players into `/client-lib/ctrls`
* move `formatTime` method into it's own file
* move `playerControls` into their own component
* move `progress` into it's own component
* move `replayView` into it's own component
* added `lib/get-soundcloud-url`
* added `lib/get-player` to determine which player is used based on the passed url
* added `audio-fft@0.1.3` dependency to help with visualizations
* added `audiosource@2.0.5` dependency to help with audio playback

- fixes #443
- fixes #442

### testing
sample frequency view
min-vid-playing-frequency-vis
the player/visualization should scale in all directions(I'm working on making the resolution better for all screens and sizes)
min-vid-playing-frequency-wide
I've changed the progress bar to peek at the bottom when there is no focus on the player so the user can
have an intrusive view of their place in the media's playback
min-vid-playing-progress-peek
sample time domain view
min-vid-playing-time-vis
User's can click on the current time to switch between time/frequency views when audio is being played back.
min-vid-switch-vis

* `mkdir media`
* put an mp3 called `mother-earth.mp3` in the dir
* put an mp4 called `seneca.mp4` in the media dir
* npm install && npm run dev
* open [localhost:8080](localhost:8080) and start launching links from the context menu
* when using audio, you can click the current time in the player to switch visualizations from time to frequency

This includes some pretty big refactoring so I'd like to get a lot of eyes on it.
cc/ @6a68 @chuckharmston @clouserw @lmorchard


6a68:
@meandavejustice Maybe it would be good to add (very small) media files to the repo? If you put them someplace out of the way, like a `test` or `local` directory, then there wouldn't be any need to copy files around. Would be good for contributors and good for QA.

app.js
```diff
@@ -27,11 +18,22 @@ window.AppData = new Proxy(defaultData, {
         window.console.error('Unable to parse l10n strings: ', ex);
       }
     } else obj[prop] = value;
+
+    if (prop === 'src') {
```
6a68: Maybe worth checking if the old 'src' matches the new 'src', and only emitting if it's actually a change

client-lib/ctrls/audio-ctrl.js
```diff
@@ -0,0 +1,91 @@
...
+  get duration() {
+    return this.audio.time().total;
+  }
+
+  play(t) {
+    this.audio.play(t);
+  }
+
+  pause() {
+    // TODO(DJ) ensure that source has loaded correctly before
```
6a68: Probably good not to land unresolved TODOs. Either fix or file a bug and add a bug number to the comment

client-lib/ctrls/audio-ctrl.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  unmute() {
+    this.volume = this.prevVolume;
+  }
+
+  remove() {
+    // disconnect nodes and stop playback
+    // remove any listeners
+    // cleanup excess nodes
```
6a68: not totally sure this comment is 100% truthful, the function calls seem a little different

client-lib/ctrls/video-ctrl.js
```diff
@@ -0,0 +1,47 @@
...
+    this.video.muted = true;
+  }
+
+  unmute() {
+    this.video.muted = false;
+  }
+
+  remove() {
+    // listeners removed in removeVideoListeners method in
+    // `components/player-view.js` via `componentWillUnmount`
```
6a68: what does this mean? does this class not need the remove method at all?

client-lib/ctrls/yt-ctrl.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  play() {
+    window.YTPlayer.playVideo();
+  }
+
+  pause() {
+    // safeguard in case pause is called when the iframe
+    // no longer exists.
+    if (window.YTPlayer.a.contentWindow) window.YTPlayer.pauseVideo();
```
6a68: if the iframe is gone, will `window.YTPlayer.a` exist?

client-lib/defaults.js
```diff
@@ -0,0 +1,19 @@
...
+  loaded: false,
+  error: false,
+  muted: false,
+  exited: false,
+  time: '0:00 / 0:00',
+  currentTime: 0,
+  duration: 0,
+  progress: 0.001, // force progress element to start out empty
+  playing: false,
+  volume: '0.5',
```
6a68: I noticed in the debugger that React seems to convert this into a float. should it be set as a float here?

client-lib/nsa.js
```diff
@@ -0,0 +1,210 @@
...
+
+const YtCtrl = require('./ctrls/yt-ctrl');
+const AudioCtrl = require('./ctrls/audio-ctrl');
+const VideoCtrl = require('./ctrls/video-ctrl');
+
+let currentStep, playerMap = {}; // eslint-disable-line no-unused-vars
+
+function step() {
+  let currentTime = 0;
+  let progress = window.AppData.progress;
```
6a68: Suggestion: the `window.AppData.foo` pattern recurs in this file a _lot_. What about wrapping the `window.AppData` proxy object in a module? This change would way simplify unit testing code that touches AppData (just override require to pass in a sinon spy), and this file would become cleaner, too:
```js
const d = require('./app-data');

// instead of let progress = window.AppData.progress, you could do:
let progress = d.progress;
```

client-lib/nsa.js
```diff
@@ -0,0 +1,210 @@
...
+    time: `${formatTime(0)} / ${formatTime(window.AppData.duration)}`,
+    progress: 0.001
+  });
+})
+
+emitter.on('init', (opts) => {
+  sendMetricsEvent('player_view', 'init');
+
+  if (window.AppData.player === 'audio') {
+    playerMap['audio'] = new AudioCtrl(opts);
```
6a68: What exactly are the responsibilities of this class? It seems like it not only listens for metrics-worthy events (lol nsa.js) but also does critical app setup.

Since you're using an emitter, maybe just record the metrics events in this file, and have another file that does the player setup, called something suggestive like `main.js` or whatever.

6a68: Related suggestion: seems like you could remove the last couple of `window.AppData` get/set calls from the components and just use emitters there instead. Looks like just a couple of instances in the player and progress components. I think it'd make more sense to have a hard rule than to sometimes use emitters and sometimes update state directly. Plus, again, being consistent about emitter usage simplifies unit testing components: you'd just mock out the emitter library.

client-lib/nsa.js
```diff
@@ -0,0 +1,210 @@
...
+
+  } else {
+    playerMap['video'] = new VideoCtrl(opts);
+  }
+});
+
+emitter.on('play', (opts) => {
+  sendMetricsEvent('player_view', 'play');
+  // TODO(DJ): this is a special case to handle audio replays this is
+  // a work around for a bug in the audiosource dependency
```
6a68: this TODO needs a bug number

client-lib/nsa.js
```diff
@@ -0,0 +1,210 @@
...
+  });
+});
+
+emitter.on('set-time', (opts) => {
+  playerMap[window.AppData.player].time = opts.value;
+
+  // if we are paused force the ui to update
+  if (!window.AppData.playing) {
+    window.AppData = deepAssign(window.AppData, {
+      time: `${formatTime(opts.value)} / ${formatTime(window.AppData.duration)}`,
```
6a68: l10n nit: this might need to be reversed in rtl languages. maybe file a good first bug if it's not an obvious, easy fix

6a68: hmm. it might also make sense to split this time thing into 'elapsed time' and 'total time'. that would let you do the locale-specific display stuff in the display code, where it belongs, and not in the data management code.

client-lib/nsa.js
```diff
@@ -0,0 +1,210 @@
...
+  window.AppData.minimized = true;
+});
+
+emitter.on('maximize', () => {
+  sendMetricsEvent(getView(), 'maximize');
+  sendToAddon({action: 'maximize'});
+  window.AppData.minimized = false;
+});
+
+emitter.on('send-to-tab', () => {
```
6a68: I'd suggest moving the send-to-tab listener into the sendToAddon library. keep the related stuff together.

6a68: or, at least create another listener that does the sendToAddon bit. might require another event, since it looks like currentTime gets updated just before the addon is notified.

components/error-view.js
```diff
@@ -27,14 +31,16 @@ module.exports = React.createClass({
...
     sendMetricsEvent('error_view', 'render');
-  },
-  render: function() {
+  }
+
+  render() {
     return (
-        <div className='error' onMouseEnter={this.enterView} onMouseLeave={this.leaveView}>
-          <div className={cn('controls', 'drag', {hidden: !this.state.hovered, minimized: this.props.minimized})}>
+        <div className='error' onMouseEnter={this.enterView.bind(this)} onMouseLeave={this.leaveView.bind(this)}>
```
6a68: I'm surprised you need `this` binding. does react not take care of this?

components/player-view.js
```diff
@@ -1,318 +1,137 @@
...
 const deepAssign = require('deep-assign');
+const emitter = require('../client-lib/emitter');
+const PlayerControls = require('../components/player-controls');
+const ReplayView = require('../components/replay-view');
+const Progress = require('../components/progress');
+
+module.exports = class Player extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {hovered: false};
```
6a68: This seems to occur everywhere. Any reason it couldn't be pulled up into a parent base class?

components/player-view.js
```diff
@@ -1,318 +1,137 @@
...
-  componentDidUpdate: function(prevProps, prevState) {
+  componentDidUpdate(prevProps, prevState) {
     if (prevProps.src !== this.props.src) {
-      this.attachVideoListeners();
+      clearTimeout(this.loadingTimeout);
+      this.loadingTimeout = setTimeout(() => {
+        deepAssign(window.AppData, {
+          error: true
+        });
+      }, 30000); // 30 second timeout
```
6a68: I wonder if some kind of intermediate "working on it..." state should be shown after 5-10 seconds. 30 seconds is a really long time to wait on the web

components/player-view.js
```diff
@@ -1,318 +1,137 @@
...
-          else if (ev.data === PAUSED && this.props.playing) this.pause();
-        },
-        onError: (err) => {
-          const url = `https://youtube.com/watch?v=${this.props.id}`;
-          window.AppData.error = `There was an error loading your video from ${url}`;
-          console.error('Error: ytCtrl: ', err); // eslint-disable-line no-console
-        }
+    if (this.props.player === 'youtube') {
+      emitter.emit('init', {
+        onLoaded: this.onLoaded.bind(this)
```
6a68: Using an emitter to pass around a function reference? This seems weird to me.

Could you fire a new event instead and wire up the function directly inside the new event's listener?

components/player-view.js
```diff
@@ -1,318 +1,137 @@
...
-    sendMetricsEvent('player_view', 'replay');
-    this.refs.video.currentTime = 0;
-    this.step(); // step once to set currentTime of window.AppData and progress
-    this.play();
-  },
-  close: function() {
-    sendMetricsEvent('player_view', 'close');
+  onError() {
+    if (!this.props.src) return;
+    window.AppData.error = true;
```
6a68: Fire error event instead of setting AppData here?

lib/get-player.js
```diff
@@ -0,0 +1,17 @@
...
+  if (opts.domain === 'soundcloud.com') player = 'audio';
+  if (isAudio(opts.src)) player = 'audio';
+
+  return player;
+}
+
+function isAudio(src) {
+  const isSoundcloud = new RegExp('^(https?:)?//api.soundcloud.com\/tracks\/[0-9]+\/stream')
+        .exec(src);
+  const isAudio = new RegExp('^(https?:)?//*.+(.mp3|.opus|.weba|.ogg|.wav|.flac)')
```
6a68: Definitely should DRY up these recurring RegExps. Maybe put inside a 'supported-formats' lib or something, just so you don't have to remember to update multiple spots if you add / remove formats.

fwiw, the regexp literal is compact and pretty nice, except you'd have to escape the // :

```js
const isAudio = /^(https?:)?\/\/*.+(.mp3|.opus|.weba|.ogg|.wav|.flac)/
```

lib/get-soundcloud-url.js
```diff
@@ -0,0 +1,23 @@
+const Request = require('sdk/request').Request;
+
+module.exports = getSoundcloudUrl;
+
+const clientId = 'c23df04aed9a788cd31fd5b100f22a7a';
+
+function getSoundcloudUrl(videoUrl, cb) {
+  Request({
+    url: 'https://api.soundcloud.com/resolve.json?client_id=' + clientId + '&url=' + videoUrl,
```
6a68: This feels like it should be a template string constant defined on the class

locales/en-US/addon.properties
```diff
@@ -6,8 +6,11 @@ tooltip_send_to_tab= Send to tab
...
+tooltip_switch_visual= Click to switch visualization
+media_type_audio= audio
+media_type_video= video
 error_msg= Something's gone wrong with this video. Try again later.
 error_link= Open in new tab
 # LOCALIZATION NOTE: placeholder is domain name
-loading_msg= Loading video from %s
-content_menu_msg= Send to mini player
\ No newline at end of file
+loading_msg= Loading %s from %s
```
6a68: you'll need to update the localization note to make clear which placeholder is which

6a68:
@meandavejustice Aight, did a first pass, made some suggestions. I'm fine to defer most of them until later, provided a bug is filed.

I'll give the other reviewers a chance to weigh in. Ping me when you'd like me to do a final nitpicky pass after you've made changes and/or filed followup bugs.

client-lib/ctrls/yt-ctrl.js
meandavejustice: yeah, this is a weird edge case where react removes it, but the reference seems to still exist on the YTPlayer object

client-lib/defaults.js
meandavejustice: I should probably go set this to a float everywhere, I think it was an oversight at some point and I ended up converting them all to strings for consistency. Will fix

client-lib/nsa.js
meandavejustice: mm I like that alot, not sure why I didn't think of that before :)

components/error-view.js
meandavejustice: Not when using es6 classes, https://facebook.github.io/react/docs/handling-events.html

components/loading-view.js
```diff
@@ -2,20 +2,24 @@ const React = require('react');
 const cn = require('classnames');
 const GeneralControls = require('./general-controls');
 
-module.exports = React.createClass({
-  getInitialState: function() {
-    return {hovered: false};
-  },
-  enterView: function() {
+module.exports = class ErrorView extends React.Component {
```
meandavejustice: Whoops

client-lib/ctrls/video-ctrl.js
meandavejustice: Yeah, removing now

client-lib/ctrls/audio-ctrl.js
meandavejustice: leftover comment from when I was planning this out

components/player-view.js
meandavejustice: onLoaded clears the timeout and calls the onError method, so It needs to be on the component.

lib/get-player.js
meandavejustice: #525

client-lib/ctrls/audio-ctrl.js
meandavejustice: https://github.com/meandavejustice/audiosource/issues/5

client-lib/nsa.js
meandavejustice: https://github.com/meandavejustice/audiosource/issues/6

meandavejustice:
@6a68 regarding https://github.com/meandavejustice/min-vid/pull/524#issuecomment-266839606

I think I would rather add a task to create the `media` dir and download some files from a server. I'd really hate to bloat the repo with extra files, it could really slow down pulling/cloning on poor connections

app.js
meandavejustice: There shouldn't be a case where src is updated with the same value

app.js
```diff
@@ -27,11 +18,22 @@ window.AppData = new Proxy(defaultData, {
         window.console.error('Unable to parse l10n strings: ', ex);
       }
     } else obj[prop] = value;
```
meandavejustice: Hmm, I don't think I'm very concerned about this type of bug, we only send data down in launch-video, i think It would be excessive to add these checks/validation

client-lib/nsa.js
meandavejustice: created issue https://github.com/meandavejustice/min-vid/issues/526

components/player-view.js
meandavejustice: bumped down to 20 seconds for now and opened https://github.com/meandavejustice/min-vid/issues/527

meandavejustice:
@6a68 You mind taking another look at this?

6a68:
I just asked again in the status mtg to get another set of eyes on this. Letting it sit for a little bit to see if anyone makes time to help out.

If nothing major comes up, I'll plan to do my second & final pass this afternoon.

6a68:
@meandavejustice I have a review in process; finishing it is at the top of my todo list for tomorrow. Couple questions for now:

When I right-click the 'bad' souncloud link, there's no minvid context menu option. Is that intentional, or a bug?

Also, I can't get the 'good' soundcloud test link to work. It looks like `client_id=undefined` in the first error in this screenshot:

screen shot 2016-12-19 at 5 02 13 pm

Am I missing a setup step? I just did `npm install && npm run dev` with dev edition running.

meandavejustice:
@6a68 I've just pushed a fix for the soundcloud issue. I had mistyped the object I was accessing the client id from in `lib/get-soundcloud-url`.

As for the context menu not working on "bad soundcloud url", this is intended. The regex that checks which urls to apply the `send-to-player` option do not match. I've updated the `index.html` to add a separate option for a soundcloud track which will error on fetch, and updated the link text for the "no context menu" option.

lib/context-menu-handlers.js
```diff
@@ -17,6 +18,23 @@ const contentScriptVimeoCase = `self.on('click', function (node, data) {
   self.postMessage('https://vimeo.com/' + node.href);
 });`
 
+const contextMenuMediaContentScript = `
+self.on('click', function (node, data) {
+  self.postMessage({
+    src: node.href || node.src,
+    url: window.location.href,
+    domain: new URL(node.href || node.src).host
```
6a68: The `URL` constructor throws if the string isn't a valid URL. Could you wrap this in a try-catch, and not send the postmessage if it throws?

lib/context-menu-handlers.js
```diff
@@ -17,6 +18,23 @@ const contentScriptVimeoCase = `self.on('click', function (node, data) {
...
+  });
+});`;
+
+const contextMenuSoundcloudContentScript = `
+self.on('click', function (node) {
+  self.postMessage({
+    url: node.href,
+    domain: new URL(node.href).host
```
6a68: Same thing here, wrap in a try/catch, and don't send unless it's valid.

lib/get-soundcloud-url.js
```diff
@@ -0,0 +1,22 @@
...
+  Request({
+    url: `https://api.soundcloud.com/resolve.json?client_id=${clientId}&url=${videoUrl}`,
+    onComplete: function (resp) {
+      if (!resp.json) {
+        cb('Could not connect to soundcloud.com');
+        return;
+      }
+
+      if (resp.json.kind !== 'track' || !resp.json.streamable) {
+        cb('not track or not streamable');
```
6a68: For telling the user what went wrong, might be useful to split this error to distinguish between the 'not track' and 'not streamable' cases? Maybe a good first bug?

package.json
```diff
@@ -20,13 +20,16 @@
   "bugs": {
     "url": "https://github.com/meandavejustice/min-vid/issues"
   },
+  "config": {
```
6a68: Not a change you made in this PR, but the version `0.3.3b` should be `0.3.3-b` to follow semver.

client-lib/format-time.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function(seconds) {
+  const secondsInDay = 24 * 60 * 60;
+  const days = Math.floor(seconds / secondsInDay);
+  seconds = seconds % secondsInDay;
+  const utc = new Date(seconds * 1000);
+  return `${days}:${utc.toLocaleTimeString('en-US', {hour12: false, timeZone: 'UTC'})}`
```
meandavejustice: https://github.com/meandavejustice/min-vid/issues/526

client-lib/app-data.js
```diff
@@ -0,0 +1,49 @@
...
+  id: '',
+  src: '',
+  url: '', // only used for <audio>, <video> tags, and soundcloud
+  domain: '',
+  minimized: false,
+  loaded: false,
+  error: false,
+  muted: false,
+  exited: false,
+  time: '0:00 / 0:00',
```
meandavejustice: https://github.com/meandavejustice/min-vid/issues/526

lib/context-menu-handlers.js
meandavejustice: Good catch, will do

lib/context-menu-handlers.js
meandavejustice: 👍

lib/get-soundcloud-url.js
meandavejustice: This currently only sets the error to TRUE when sent down to the client, it's more or less a placeholder until we add more specific error messaging. I've created https://github.com/meandavejustice/min-vid/issues/532 to track these types of instances

package.json
meandavejustice: 👍

meandavejustice:
@6a68 Thanks for your review

>Bug 1: when audio is loading, close button doesn't work.
Woah, gunna take care of this one, the underlying library that I had written a while back really needs a refresh, will be eager to do that soon. Will take care of this in our code for now.

> Bug 2: when audio is loading, send to tab doesn't close the window.
Pretty much, same as above :)

I think I've address the rest of your comments. Let's talk in the morning to see if there is anything else we should try and get in before the next release.

components/error-view.js
```diff
@@ -4,17 +4,21 @@ const sendToAddon = require('../client-lib/send-to-addon');
...
 
-module.exports = React.createClass({
-  getInitialState: function() {
-    return {hovered: false};
-  },
-  enterView: function() {
+module.exports = class ErrorView extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {hovered: false};
```
meandavejustice: It has no effect on the actual dom's state. It's just initializing the state object with a value. The same way I would initialize a variable representing hover state in a vanilla js project.

``` js
var textHovered = false;
document.querySelector('.text').addEventListener('mouseenter', (ev) => textHovered = true);
document.querySelector('.text').addEventListener('mouseleave', (ev) => textHovered = false);
```

components/player-view.js
meandavejustice: `onLoaded` is the same regardless since it is just the method right here in `player-view` component. However, I need to be able to call it at different times depending on which `ctrl` is being used.
For instance. in `audio-ctrl` I need to load the underlying file before calling `onLoaded` in that file.

With YouTube I need to call it withing the `init` method because that is where we initialize the playerMap objects so that we can control playback in one place for all events(play, pause, etc..)

meandavejustice:
can't reply to https://github.com/meandavejustice/min-vid/pull/524#discussion_r93340515 for some reason so:
>Sure, I was more curious about why you have to manually initialized hovered to false. Is this just to clear out hover state when switching between views, like from loading to player view?

Variables for states like "hover" should never be initialized with true, since it is highly unlikely that the user is hovering over the element on initialization, and they are set to hover on the `mouseenter/over` events and removed on `mouseleave`

6a68:
> Variables for states like "hover" should never be initialized with true, since it is highly unlikely that the user is hovering over the element on initialization, and they are set to hover on the mouseenter/over events and removed on mouseleave

Right, makes sense. I wasn't thinking that the data belonged only to the specific component, not to the global data store.

6a68:
@meandavejustice Looks like those major bugs were fixed. Thanks for filing followups for the other little issues. Merging :tada:

Issue #331, opened 2016-09-28
Support any HTML5 video


6a68:
We have legal approval to support any video element if:
- the page that hosts the video is loaded
- preroll ads are shown in the player whenever present

Need to figure out the best way to clone a playing video element into the min vid panel. Maybe Document.importNode?

padenot:
We should try `HTMLMediaElement.captureStream`. Come to #media on mozilla's irc if you want to chat about it (same nick, Paris timezone). 

This call allows forwarding an HTMLMediaElement to another one (using a MediaStream to connect the two), sharing all the resources behind the scene (decoders, frames, etc.). It's kind of standard (there is a spec, but we don't respect it to the letter), maybe you need a `moz` prefix.

In any case, this is awesome, and we (people doing media work in gecko) will be happy to talk if you need or want to know anything.

Links:
- Doc: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/captureStream (don't trust the compat table, it's been in gecko for ever, prefixed)
- Spec: https://w3c.github.io/html-media-capture/#widl-HTMLMediaElement-captureStream-MediaStream

6a68:
@padenot Sounds great! I'll ping you tomorrow. How late are you around in the evenings? Does 6pm Paris time work?

padenot:
Yep, should be alright.

meandavejustice:
After looking over @padenot's comments in #media this morning to @6a68, I tried hacking together a version of this with the sdk. I assumed there would be some permissions issues, and there was.

Here is a list of steps describing what we want to achieve:
* launch a pagemod (served from resource://) into a page with a tag
* click a button on the page which will call captureStream returning a "mediasource" object
* open a new window with our own tag and controls. (served from resource://)
* set the tag in our new window's 'mozSrcObject' property to the "mediasource" object retrieved in the pagemod.
* launch a pagemod (served from resource://) into a page with a tag

The main problem here is getting the `mediastream` object out of the page/pagemod and into the new window.

My next steps are digging into [Components.utils.Sandbox](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Language_Bindings/Components.utils.Sandbox#options) and [nsIPrincipal](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIPrincipal).

[...]

PR #529, opened 2016-12-20
POC: use swapBrowsers and fullscreen machinery for the mini-window

nickolay:
This is a proof-of-concept for an alternative implementation of the mini-player that could be used to solve #331. I don't know if this has been considered before (sorry if it has).

The basic idea is that instead of trying to rip the video out its host page and display in a separate floating window, we could instead
* move the whole page to a separate window (just like "Move to New Window" in the tab strip's context menu does)
* and then make the video -- but not the mini-player window -- "fullscreen".

The latter can be achieved using a preference made for the B2G emulator.

This removes the delay when switching to minivid and any kind of HTML5 video (that can play full-screen) can play in the mini-player, including EME-protected videos (Netflix), which even captureStream (discussed in #331) will not handle.

I used browser.xul in this PoC, but a real implementation could use simpler chrome with just a xul:browser. It's a massive amount of work, but I think it's worth the result.

What do you think?

(To check this out, open a YouTube video, then click the minivid overlay -- all other functionality, including the UI is not implemented. I tested on the current Nightly)

### DIFF
--- a/data/icon-overlay.js
+++ b/data/icon-overlay.js
@@ -78,6 +78,7 @@ function ytHomePageHandler(el) {
   const tmp = getTemplate();
   tmp.addEventListener('click', function(ev) {
     evNoop(ev);
+
     self.port.emit('launch', {
       url: 'https://youtube.com' + url,
       domain: 'youtube.com'
@@ -94,8 +95,8 @@ function ytWatchElementHandler(el) {
   tmp.addEventListener('click', function(ev) {
     evNoop(ev);
     const videoEl = document.querySelector('video');
-    videoEl.pause();
     closeFullscreen();
+    //videoEl.pause(); // no need to pause the video, since it continues playing in the new window.
     self.port.emit('launch', {
       url: window.location.href,
       domain: 'youtube.com',
--- a/lib/window-utils.js
+++ b/lib/window-utils.js
@@ -28,21 +28,51 @@ let commandPollTimer;
 // waits till the window is ready, then calls callbacks.
 function whenReady(cb) {
   // TODO: instead of setting timeout for each callback, just poll, then call all callbacks.
+  if (mvWindow && mvWindow.gBrowserInit && mvWindow.gBrowserInit.delayedStartupFinished) return cb();
+  /*
   if (mvWindow &&
       'AppData' in mvWindow.wrappedJSObject &&
       'YT' in mvWindow.wrappedJSObject &&
       'PlayerState' in mvWindow.wrappedJSObject.YT) return cb();
+  */
   setTimeout(() => { whenReady(cb) }, 25);
 }
 
 // I can't get frame scripts working, so instead we just set global state directly in react. fml
 function send(eventName, msg) {
   whenReady(() => {
+    if (eventName === 'set-video') {
+      // TBD: this is called twice for some reason
+
+      // instead of making the window widget full-screen, just make the video
+      // element cover the whole window when fullscreen is requested
+      Services.prefs.setBoolPref("full-screen-api.ignore-widgets", true);
+
+      // request fullscreen from the context of the page, but with chrome
+      // privileges -- to avoid the restriction that fullscreen must be
+      // requested directly in response to user action.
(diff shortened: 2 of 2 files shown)


meandavejustice:
@nickolay This is a really smart idea. Thanks for putting together a proof of concept. This seems like it may be a reasonable way to achieve what we want. I'm going to play with this a bit, and try and think about how this will work with audio/queues and future plans today and update you with any questions.

cc/ @6a68 check this out!

6a68:
@nickolay Hi, thanks for the pull request. This is a very exciting proposal :+1:

The fullscreen approach has been discussed, but not on a public mailing list; thank you very much for thinking this through and putting a POC patch together. I've pinged Xidorn to get platform feedback on whether this approach could work. He originally raised concerns about some dangerous fullscreen edge cases, which your patch seems to avoid (not moving a fullscreened doc between windows; and, ensuring the pref stayed set while the minvid player was open). I'm not sure if there are other problems that he didn't mention at the time. Hopefully he will add some feedback soon.

How would we support regular fullscreen in Firefox with this patch? If a user played a video in minvid, then tried to fullscreen a different window, would it work if we intercepted the fullscreen request, closed the minvid player, unset the pref, and then allowed the fullscreen request to proceed?

Also, do you happen to know if the `full-screen-api.ignore-widgets` pref, or the code that it toggles, is scheduled for removal? I've seen some recent threads about removing B2G code from the tree, and B2G emulator support seems like an easy target for such cleanup work.

All in all, this is an exciting possibility for min-vid. If platform thinks this could work, would you be interested in taking the patch the rest of the way? If so, we'd be happy to add you to the contributors list on the test pilot site, and send you some contributor swag as well. And I could give you a clear list of items to consider when finalizing the patch. If not, no worries :-)

nickolay:
The use of a b2g-specific pref is just a kludge I used to avoid any changes to Firefox itself. Ideally this would implemented as a platform API (e.g. `video.mozRequestFullScreen({notReallyFullScreen: true})`).

The value of the pref is only important until the requestFullscreen() call has been processed (it's checked [here](http://searchfox.org/mozilla-central/rev/46ded325e8f60b3d4c731b27c31d383911056a38/dom/base/nsGlobalWindow.cpp#6562)). I didn't include the code to reset the pref back to its default value after the fullscreen request is processed, but it's certainly doable.

A lot of work needs to be done before this could be adopted in mainline min-vid, and I won't be able to finalize it; this is just an idea I wanted to share.

6a68:
I followed up with Xidorn on IRC, and he thought this could indeed work, though it would break fullscreen for sure on non-e10s users. We can show some kind of messaging to the user when that happens, though. For e10s, we should be able to intercept the fullscreen request, close the minvid window, unset the pref, then go fullscreen with the requested tab.

So, since this could work, and supporting arbitrary html5 videos was the whole point of this thing to begin with. I'm going to follow up with Test Pilot product to make sure there's room in the schedule to ship this change. If the schedule allows, Dave or I will get a branch going and try to ship this early in Q1.

@nickolay Seriously, great work here. You found hacks around a couple of issues that sounded blockerish when I initially asked platform about this approach. As a token of our gratitude, would you like some Test Pilot stickers? If so, ping me with your mailing address (_6a68 on IRC or 6a68 at mozilla dot com). I have to order a new batch, so it might take a couple weeks.

Going to close this POC pull request. Thanks again 👍

nickolay:
Cool, glad I could help!

Re breaking fullscreen: I still don't understand when or why it would break. The pref can be unset as soon as the mini-player is initialized; after that the situation is not much different from two full-screen windows on two monitors.

FWIW, the following steps seem to work fine for me in non-e10s with the PoC code:
* open a youtube video, open in min-vid
* go to about:config and reset the pref (this would need to be done automatically, I just didn't implement it)
* open another youtube window and make it fullscreen

(By the way I didn't notice that I tested the PoC in non-e10s profile; running it under e10s fails to make the video mini-player go fake-fullscreen, with "Request for fullscreen was denied because requesting element is not in the currently focused tab." message in the console)

PR #530, opened 2016-12-20
Parse cc option from youtube watch page and pass to min vid

### LINKED ISSUE #522: Closed Captions/Subtitles are displayed in min-Vid player even if they are set to OFF on YouTube
**[Affected versions]:**
-Firefox 45.0 and up
-Min-Vid 0.3.3b

**[Affected Platforms]:**
-All Windows

**[Prerequisites]:**
-Latest Min-Vid add-on version (0.3.3b) was previously installed on a clean profile.
-Need to have an YouTube video that has Closed Captions/Subtitles, not auto-translate. ( eg. https://www.youtube.com/watch?v=P3Ew8dCUonE )

**[Steps to reproduce]:**
2. Navigate to YouTube.
3. Search for a video that has subtitles (eg. https://www.youtube.com/watch?v=P3Ew8dCUonE ).
4. Play the video on YouTube and set the Closed Captions/Subtitles to OFF.
5. Open the video in Min-Vid player.

**[Expected result]:**
- Closed Captions/Subtitles are OFF in Min-Vid player.

**[Actual result]:**
-  Closed Captions/Subtitles are displayed in Min-Vid player.

**[Notes]:**
-[Here](https://drive.google.com/open?id=0B6QogV2AJFKiaUlzWGx1Q2xEcHc) is a recording of the issue.
-Closed Captions/Subtitles should be displayed in Min-Vid player only if they are set to ON for the played video on YouTube.

meandavejustice:
- fixes #522 
- refs #68

### DIFF
--- a/lib/get-youtube-url.js
+++ b/lib/get-youtube-url.js
@@ -3,7 +3,9 @@ const qs = require('sdk/querystring');
 
 module.exports = getYouTubeUrl;
 
-function getYouTubeUrl(videoId, cb, time) {
+function getYouTubeUrl(videoId, cb, time, cc) {
+  if (!cc) cc = false;
+
   const params = qs.stringify({
     autoplay: time ? 1 : 0,
     modestbranding: 1,
@@ -11,6 +13,7 @@ function getYouTubeUrl(videoId, cb, time) {
     disablekb: 0,
     enablejsapi: 1,
     fs: 0,
+    cc_load_policy: +cc, // force boolean to number
     iv_load_policy: 3,
     loop: 0,
     rel: 0,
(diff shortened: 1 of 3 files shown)


lib/get-youtube-url.js
```diff
@@ -3,14 +3,17 @@ const qs = require('sdk/querystring');
...
+  if (!cc) cc = false;
+
   const params = qs.stringify({
     autoplay: time ? 1 : 0,
     modestbranding: 1,
     controls: 0,
     disablekb: 0,
     enablejsapi: 1,
     fs: 0,
+    cc_load_policy: +cc, // force boolean to number
```
6a68: Seems weird to use `Boolean(foo)` instead of `!!foo`, but then use `+foo` instead of `Number(foo)`.

I personally prefer the shorthand version, but I care more that it's consistent. Your call on what approach to pick.

meandavejustice: I agree, I'll update this to the shorthand

6a68: Cool. Feel free to merge this yourself when that's done :+1:

PR #536, opened 2016-12-22
Fix launch issue for some soundcloud links

### LINKED ISSUE #534: Min-Vid player is not working when tracks from Soundcloud searches are sent to it
[Affected versions]:
- Nightly 53.0a1
- Min-Vid 0.3.3c (built on 12/22/2016)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Min-Vid add-on version (0.3.3c) was previously installed on a clean profile.

[Steps to reproduce]:
2. Navigate to https://soundcloud.com/ and search for a keyword.
3. Right click on a song name or thumbnail and send the track to the mini player.
4. Observe the top left part of the screen.

[Expected result]:
- Min-Vid is opened and song is paused.

[Actual result]:
- Min-Vid window is white and unresponsive.

[Notes]:
min-vid window is white after sending a soundcloud track to it

- Console output:
capture

meandavejustice:
- context menu sometimes sends the exact target instead of the link
 target added some code to handle these cases for soundcloud.
- fixes #534

### DIFF
--- a/lib/context-menu-handlers.js
+++ b/lib/context-menu-handlers.js
@@ -20,7 +20,6 @@ const contentScriptVimeoCase = `self.on('click', function (node, data) {
 
 const contextMenuMediaContentScript = `
 self.on('click', function (node, data) {
-  console.log('contextMenuMediacontentscript', self, node, data);
   let host = '';
   try {
     host = new URL(node.href || node.src).host;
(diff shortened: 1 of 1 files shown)


lib/context-menu-handlers.js
```diff
@@ -33,16 +32,24 @@ self.on('click', function (node, data) {
...
-self.on('click', function (node) {
-  self.postMessage({
-    url: node.href,
-    domain: 'soundcloud.com'
-  });
+self.on('click', function (node, data) {
+  // if the target element isn't the link we
+  // are looking for, recurse until we find it.
+  function setParentNode(node) {
+    if (node.href) {
```
6a68: How about `if (node && node.href)`? This would guard against the case where something goes horribly wrong, and you wind up going all the way to the document's `null` parentNode, try to access a property on null, and trigger a TypeError

6a68: The functin name is kinda funny, now that I'm looking at it. Might be nicer as a while loop. Just a suggestion, though. Fine to leave as-is.

6a68: You're good to land this after addressing the null `node` check.

PR #588, opened 2017-01-27
Integrate testpilot-metrics package

meandavejustice:
- add support for ga
- add `testpilot-metrics@2.0.1`

### DIFF
--- a/lib/send-metrics-data.js
+++ b/lib/send-metrics-data.js
@@ -1,7 +1,14 @@
-/* global Services */
+const Metrics = require('testpilot-metrics');
+const store = require('sdk/simple-storage').storage;
+const manifest = require('../package.json');
 
-const {Cu} = require('chrome');
-Cu.import('resource://gre/modules/Services.jsm');
+const { sendEvent } = new Metrics({
+  id: require('sdk/self').id,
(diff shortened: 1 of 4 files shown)


lib/send-metrics-data.js
```diff
@@ -1,7 +1,14 @@
-/* global Services */
+const Metrics = require('testpilot-metrics');
+const store = require('sdk/simple-storage').storage;
+const manifest = require('../package.json');
 
-const {Cu} = require('chrome');
-Cu.import('resource://gre/modules/Services.jsm');
+const { sendEvent } = new Metrics({
+  id: manifest.name,
```
6a68: I was wrong - turns out you do need the '@' symbol. I think this is the nicest way to get it: `require('sdk/self').id`.

lib/send-metrics-data.js
```diff
@@ -1,7 +1,14 @@
...
+  type: 'sdk',
+  version: manifest.version,
```
6a68: nicer: `require('sdk/self').version`

lib/send-metrics-data.js
meandavejustice: 👍 fixing now

Issue #735, opened 2017-04-11
"Add to queue" toast message too small


youwenliang:
When a user clicks "+", the toast message display on the MinVid player is too small. The text size should be at least the same size as the tooltips.

Ideally, I propose that once a user clicks on the "+" icon, the icon should temporary change to a "✓" icon and turn back to "+" after a few seconds, so that there's also a visual feedback on the clicking area.

![001](https://cloud.githubusercontent.com/assets/8919308/24902080/a5f9627a-1edb-11e7-9483-4f3ec7092178.gif)

SoftVision-PaulOiegas:
The "Track added to queue" message is now large enough in my opinion to be easily noticed. I also verified if the message is affected in some way when setting higher resolution for the player, but it remains the same.

Issue #736, opened 2017-04-11
No hover state on several icons


youwenliang:
The "Play/Pause", "Previous/Next", and "Volume" icons should have a hover state that changes the transparency to 100% (White), just like the "close", "send to tab", "minimize" icons.

![002](https://cloud.githubusercontent.com/assets/8919308/24902244/23eaa9d2-1edc-11e7-9793-47a8b1af872d.gif)

SoftVision-PaulOiegas:
Verified this using a custom build made from the repo today (04/26/2017) and now all the buttons have 100% white hover state.

Issue #737, opened 2017-04-11
The overlay icons should stay and should have hover state


youwenliang:
I expect that the Play Now & Add to Queue icons should stay as long as I'm hovering on the video, but now it will fade away after a few seconds.

Also, when I hover on the "Play Now" icon or "Add to Queue" icon, both icons turn into 100% transparency. I expect that only the one I hover on should change to 100% transparency.

![003](https://cloud.githubusercontent.com/assets/8919308/24902401/8ad82c28-1edc-11e7-88a0-43fc45c368e4.gif)

SoftVision-CiprianMuresan:
Hi @youwenliang, thank you for reporting this, however it is a dupe of #439.

Issue #439, opened 2016-10-18
Min-Vid icon is not displayed on the video thumbnail when moving the mouse cursor, if previously the icon has faded out


Softvision-VladBacia:
[Affected versions]:
-Firefox 45.0 and up
-Min-Vid v0.3.1

[Affected Platforms]:
-All Windows

[Prerequisites]:
-Latest Min-Vid add-on version (0.3.1-dev) was previously installed on a profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Navigate to a website with videos supported by Min-Vid player (eg: YouTube, Vimeo, etc).
3. Click on a video.
4. Move the Mouse pointer over the video.
5. Keep the mouse pointer inside the video area and wait until Min-Vid icon has faded out.
6. Move the mouse pointer inside the video area.

[Expected result]:
-Min-Vid icon should be displayed when the mouse pointer is moved.

[Actual result]:
-Min-Vid icon is not displayed. When the mouse pointer is over the icon area, the link select pointer is displayed although the Min-Vid icon is not displayed.

[Additional Notes]:
-[Here](https://drive.google.com/open?id=0B6QogV2AJFKiSUthWFh2dzN3SlE) is a video of the issue.

ASCIIbreakdown:
I was aware of this "bug" when I added the fading effect. The fade effect is done with css keyframes, which are activated on hover, so if you "rehover" (move the curser out and in again) it will replay the animation. Using :hover seemed to be the best way to do it at the time, but it might be better done differently.

PR #752, opened 2017-04-14
Fix issue #735 & #736 for visual updates

youwenliang:
Fix issue #735 & #736 for visual updates


data/panel.css
```diff
@@ -795,12 +801,12 @@ canvas {
     border-radius: 4px;
     padding: 5px;
 
-    font-size: 2vw;
+    font-size: 12px;
```
meandavejustice: I'd rather have a solution that scales to different sizes

meandavejustice: May have to use media queries with a minimum size

data/icon-overlay.js
```diff
@@ -240,12 +240,15 @@ function getTemplate() {
 
   containerEl.className = 'minvid__overlay__container';
   playIconEl.className = 'minvid__overlay__icon';
+  playIconEl.id = 'minvid__overlay__icon__play';
   playIconEl.title = 'Play Now';
-  addIconEl.className = 'minvid__overlay__icon__add';
+  addIconEl.className = 'minvid__overlay__icon';
+  addIconEl.id = 'minvid__overlay__icon__add';
```
meandavejustice: Neither of the issues linked reference overlay icon, so I'm not sure what these changes are addressing.

data/panel.css
youwenliang: Agree, that was a temporary test to see the font-size differences.

data/icon-overlay.js
youwenliang: This was for the individual icon to have hover state, check the Gif below:
apr-15-2017 00-30-12

meandavejustice: Nice! Is there an issue for this one?

youwenliang: I think it was part of #737 but was considered as a duplicate.

meandavejustice: So does this take care of #439?

youwenliang: Not quite, I removed the fade effect so as long as the mouse is hovered on a video to overlay icons won't fade away. 

However, the description in #439 suggests handling the idle situation where the overlay icons still fade away but will reappear when the mouse is moving (without re-hovering).

meandavejustice:
@youwenliang happy to merge once the conflict is resolved.

PR #768, opened 2017-04-19
Generic Queue UI Text

### LINKED ISSUE #767: Wording in queue UI should be more generic
Some of the wording in the new Queue UI is geared more towards audio, where it should probably be more generic. These are purely cosmetic changes, but could prevent some confusion around what is going on in the extension.

Some changes might include
"*Track* added to queue" => "*Item* added to queue"
"There are no more *songs* in the queue." => "There are no more *items* in the queue."

Also, in relation to #750, [ReplayView](https://github.com/meandavejustice/min-vid/blob/master/components/replay-view.js) has a static string, when it should have a localised string.

proman21:
Fixes #767


meandavejustice:
cc/ @flodolo

locales/en-US/addon.properties
```diff
@@ -18,7 +18,8 @@ add_to_queue= Add to queue
 clear= Clear
 history= History
 play_queue= Play Queue
-track_added_notification= Track added to queue
+track_added_notification= Item added to queue
```
flodolo: You need to use a new ID here, otherwise tools will maintain the existing translation
This is valid also for add-ons https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

meandavejustice: @proman21 I would suggest `item_added_notification`, but then this will need to be updated in `lib/get-locale-strings.js`

meandavejustice:
@proman21 It looks like this isn't going to work without added "end_of_queue" to https://github.com/meandavejustice/min-vid/blob/master/lib/get-locale-strings.js

PR #779, opened 2017-04-25
Set max and min window sizes to prevent breakage

### LINKED ISSUE #667: There is no Min/Max resolution for the Min Vid player window
[Notes]:
- If a custom resolution of 0x0 is set on Linux, the player seems to have minimum values and it will be displayed according to them. On Windows and Mac however, the player is created with a resolution of 0x0.
- If we want to keep the idea of a background player, then the same should be done for a maximum value (when doing this we should keep in mind HiDPI monitors). As it stands now, if custom values are set that exceed the screen resolution, the player is not displayed at all.

[Affected versions]:
- Firefox 52.0 and up
- Min-Vid 0.3.4-dev (built on 3/17/2017)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Min-Vid add-on with Queue functionality was previously installed on a clean profile.

[Steps to reproduce]:
2. Go to about:addons->Extensions tab and double click the Min Vid addon.
3. Set a custom resolution of 0 height and 0 width.

[Expected result]:
- A minimum value for height and width is forced.

[Actual result]:
- The custom resolution is set.

[Additional Notes]:
min and max height and width for player window

meandavejustice:
- fixes #667

### DIFF
--- a/data/get-dimensions.js
+++ b/data/get-dimensions.js
@@ -1,0 +1,22 @@
+self.port.emit('dimensions', getDimensions());
+
+function getDimensions() {
+  let width = 0;
+  let height = 0;
+
+  if (typeof(window.innerWidth) === 'number') {
+    height = window.innerHeight;
+    width = window.innerWidth;
+  } else if (document.documentElement && (document.documentElement.clientWidth || document.documentElement.clientHeight)) {
+    height = document.documentElement.clientHeight;
+    width = document.documentElement.clientWidth;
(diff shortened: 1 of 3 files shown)


6a68:
@meandavejustice This does work, but the experience is kinda janky: if you try to type "600" in the width box in the preferences page, the "6" instantly turns into "400". Kinda annoying. But it does fix the bug, so I'll land it ^_^

PR #780, opened 2017-04-26
Hide exited view when queue items are added

### LINKED ISSUE #702: Playback does not start automatically if a track is added while Min Vid is in the Replay/Close view
[Affected versions]:
- Firefox 52.0 and up
- Min-Vid 0.3.4-dev (built on 3/29/2017)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Min-Vid add-on with Queue functionality was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites and navigate to any supported platform.
2. Send a video/song to the mini player and fast forward it almost to the end.
3. Wait for the video to end.
4. Add a different video/song using the "+" overlay button and observe the player.

[Expected result]:
- Playback starts after the item is added to Queue.

[Actual result]:
- The item is added to "Play Queue" but the Replay/Close view remains.

[Notes]:
- The issue is also reproducible if the item is added from the context menu option.
track is added while min vid is in the replay close

meandavejustice:
- fixes #702
- also made general controls visible in exited view

### DIFF
--- a/components/player-view.js
+++ b/components/player-view.js
@@ -277,7 +277,7 @@ module.exports = class Player extends React.Component {
                             <ReactTooltip id='next' effect='solid' place='right'>{this.props.strings.ttNext}</ReactTooltip>
                           </div>);
 
-    const exited = this.props.exited ? (<ReplayView {...this.props} exited={this.props.exited}
+    const exited = this.props.exited ? (<ReplayView {...this.props} nextTrack={this.nextTrack.bind(this)} exited={this.props.exited}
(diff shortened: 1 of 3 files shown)


meandavejustice:
@6a68 I guess it could go either way in this case. But I agree that autoplaying probably makes the most sense here. I'll update it to autoplay on monday.

meandavejustice:
@6a68 Autoplay now working

PR #781, opened 2017-04-26
Update visual feedback for history '+' button.

youwenliang:
A few visual updates:
- Add hover state for '+' button in the history panel.
- Remove highlighting the first item in the history panel (we only need to highlight the currently playing item in the queue panel)

Like this:
apr-26-2017 13-05-22


SoftVision-PaulOiegas:
The "History" panel no longer has the first item highlighted and the hover effect is now present for "+" button.

Issue #626, opened 2017-02-27
Playback stutters and lots of pings are sent when the "Play / Pause" button is quickly double or triple clicked


SoftVision-PaulOiegas:
[Notes]:
- When this occurs the browser performances are also affected.
- Lots of play / pause pings are sent.
- First I managed to reproduce this while having other application in full screen. You need to double click the mini player to have it in focus and to make the desired action.

[Affected versions]:
- Firefox 51.0 and up
- Min-Vid 0.3.4-dev (built on 2/27/2017)

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Latest Min-Vid add-on version (0.3.4-dev) was previously installed on a clean profile.
- Firefox is already opened with the above profile and with Browser console opened with "Net" filter on. 

[Steps to reproduce]:
1. Navigate to www.youtube.com page and send a video to the mini player.
2. Choose to "Play" the video.
3. Quickly double (harder to reproduce) or triple click on the "Play / Pause" button.
4. Observe the playback behavior and the console logs.  

[Expected result]:
- The "Play / Pause" actions are correctly made and do not affect the video playback.

[Actual result]:
- Playback shutters and lots of pings are sent when the "Play / Pause" button is quickly double or triple clicked.

[Additional Notes]:
- Haven't managed to reproduce this on Mac OS.
- Attached a screen recording of the issue: https://goo.gl/OuYyCP.

meandavejustice:
Is this still and issue in the queue branch? @SoftVision-PaulOiegas

SoftVision-PaulOiegas:
Seems so https://goo.gl/drOM0Q.

SoftVision-CiprianMuresan:
I've found that the issue is more easily reproducible if you have a lot of loaded tabs (20-30). This way the issue is also reproducible on Mac OS.

SoftVision-PaulOiegas:
The fix seems pretty ok. I have managed to reproduce this, but now I consider it a super edge case (spam holding "Space" key). I would say that we can consider it fixed. 
However, I have logged the remaining related behavior with "Replay" button separately (#836).

PR #788, opened 2017-04-28
Clean up play button behavior when last item exits

### LINKED ISSUE #782: Overlay not removed and "Play / Pause" button must be clicked 3 times for video to restart after playback end
**[Affected versions]:**
- Firefox v53.0 and up
- Min-Vid v0.3.6 (production)
- Min-Vid v0.3.6 (custom repo build)

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Latest Min-Vid add-on was previously installed on a clean profile from Test Pilot or custom repo build.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites and navigate to https://youtube.com page.
2. Hover a video thumbnail and click on "Play Now" overlay button.
3. Fast forward the video or let it end, in order to have the "Replay / Close" overlay displayed.
4. Hover the player and observe the "Play / Pause" button state.
5. Click the "Play / Pause" button and observe the player behavior.
6. Click the "Play / Pause" button several times more.

**[Expected result]:**
- Step 4: The button passes from "Pause" to "Play" state after the video playback ends and the "Replay / Close" overlay is displayed. 
- Step 5: The button passes into "Pause" state, the overlay is removed and video playback starts again.
- Step 6: The button correctly passes into the opposite state each time is clicked and the video playback is correctly paused or played.

**[Actual result]:**
- Step 4: The button wrongly remains into  "Pause" state after the video playback ends and the "Replay / Close" overlay is displayed. 
- Step 5: The button wrongly passes into "Play" state, the playback is not restarted and the overlay is not removed.
- Step 6: A second click leaves the button in the same "Play" state, only the seek bar is restarted. A second click makes the video playback to restart, but the "Replay / Close" overlay is not removed. 

**[Notes]:**
- This issue also occurs in minimized mode where there is no "Replay" button, making this issue even easier to encounter.
- This issue also occurs for SoundCloud songs but you only need to click twice the "Play / Pause" button. But the overlay still remains displayed.
- This issue **does not** occurs for Vimeo videos. The overlay is also correctly removed when "Play" button is pressed.
- Here is a screen recording of the issue: https://goo.gl/rj30tQ.

6a68:
* Toggle the play/pause button when a video ends

* If play is pressed when exited is true, then replay

Fixes #782.

-----

I verified that this fixes the issues mentioned in #782. Ran into some weirdness with the replay button getting youtube videos into the choppy state where it's manically flipping between play and pause, but I also saw that bug without this patch applied, so hopefully that's a known thing.

### DIFF
--- a/components/player-view.js
+++ b/components/player-view.js
@@ -105,7 +105,7 @@ module.exports = class Player extends React.Component {
 
   onEnded() {
     if (this.props.queue.length === 1) {
-      window.AppData.set({exited: true});
+      window.AppData.set({exited: true, playing: false});
     } else sendToAddon({action: 'track-ended'});
   }
 
@@ -284,9 +284,9 @@ module.exports = class Player extends React.Component {
                                               openQueueMenu={this.openQueueMenu.bind(this)} />
                                               <PlayerControls {...this.props} hovered={this.state.hovered} progress={this.state.progress}
                                               audio={this.audio} time={this.state.time} setTime={this.setTime.bind(this)}
-                                              closeQueueMenu={this.closeQueueMenu.bind(this)} />
+                                              replay={this.replay.bind(this)} closeQueueMenu={this.closeQueueMenu.bind(this)} />
                                               </div>) : (<MinimizedControls {...this.props} progress={this.state.progress}
-                                                         time={this.state.time} setTime={this.setTime.bind(this)} />);
+                                                         replay={this.replay.bind(this)} time={this.state.time} setTime={this.setTime.bind(this)} />);
 
     return (<div className='video-wrapper'
                  onMouseEnter={this.enterPlayer.bind(this)}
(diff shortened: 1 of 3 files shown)


components/player-view.js
```diff
@@ -105,7 +105,7 @@ module.exports = class Player extends React.Component {
 
   onEnded() {
     if (this.props.queue.length === 1) {
-      window.AppData.set({exited: true});
+      window.AppData.set({exited: true, playing: false});
```
6a68: This is necessary to flip the pause button back to a play button

meandavejustice:
@6a68 The choppy behavior with rapid clicks has been around, we should probably add some debounce handling on those, there is an issue already for that though, thanks for the patch :)

SoftVision-PaulOiegas:
> Ran into some weirdness with the replay button getting youtube videos into the choppy state where it's manically flipping between play and pause, but I also saw that bug without this patch applied, so hopefully that's a known thing.

@6a68 the issue you were mentioning #626.

PR #879, opened 2017-05-24
Metrics audit

meandavejustice:
- Update docs/metrics.md to reflect changes and add context
- send correct domain when applicable
- straighten up playlist/confirm and queue_view metric events
- stop sending duplicate play/pause events


meandavejustice:
cc/ @wresuolc @johngruen

PR #931, opened 2017-07-31
Fix topify

6a68:
I think this fixes it. Need to get some QA love though.

### DIFF
--- a/lib/topify.js
+++ b/lib/topify.js
@@ -44,12 +44,35 @@ _initPlatformCode();
 function topify(domWindow) {  // eslint-disable-line no-unused-vars
   if (unsupportedPlatform) return console.error(`Unable to topify minvid window on unsupported window toolkit ${platform}.`); // eslint-disable-line no-console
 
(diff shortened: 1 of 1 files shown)


lib/topify.js
```diff
@@ -39,16 +40,21 @@ function _initPlatformCode() {
...
 // Returns true if successful.
 function topify(domWindow) {  // eslint-disable-line no-unused-vars
   if (unsupportedPlatform) return console.error(`Unable to topify minvid window on unsupported window toolkit ${platform}.`); // eslint-disable-line no-console
-
-  const winPtrStr = getNativeHandlePtrStr(domWindow);
+  winPtrStr = getNativeHandlePtrStr(domWindow);
   if (!winPtrStr) return console.error('Unable to get native pointer for window.'); // eslint-disable-line no-console
 
+  // Topification can take up to a minute on macos. Wait a little longer,
+  // then clear the window reference, avoiding memory leaks.
```
6a68: maybe I'm being too paranoid about window reference memory leaks?

PR #977, opened 2017-08-12
Fix links in changelog

petersohn:


pdehaan:
@meandavejustice Out of curiosity, how are you generating the CHANGELOG file? We were looking to do the same thing in a couple other experiments.

meandavejustice:
@pdehaan https://www.npmjs.com/package/changelog-maker

PR #986, opened 2017-08-15
Update install.rdf version to match package.json and manifest

### LINKED ISSUE #981: Update version field in install.rdf

meandavejustice:
fixes #981


install.rdf
```diff
@@ -6,7 +6,7 @@
     <em:hasEmbeddedWebExtension>true</em:hasEmbeddedWebExtension>
     <em:type>2</em:type>
     <em:unpack>false</em:unpack>
-    <em:version>1.0.10</em:version>
+    <em:version>0.4.0</em:version>
```
pdehaan: Will this affect auto updates in any way since we're seemingly downgrading the version number from v1 to v0.4? Or should we just bump all version numbers to v2.0.0 and keep them in sync?

meandavejustice: I am a little worried about this tbh, this pr was just to get them in sync. I will need to do a version bump for a release later this week (or early next week) with the changes from https://github.com/meandavejustice/min-vid/milestone/10. We should figure this out for certain before then.

meandavejustice: @pdehaan I actually just checked and it appears that this possibly kept the addon from properly updating. When I install min vid from testpilot I get 0.3.9, the previous version. Which isn't working properly in nightly. It probably makes sense to push this fix, and the couple others and do another prod push (today if possible) to get our proper changes into nightly. I'm going to message wil about this right now.

meandavejustice:
fixed in https://github.com/meandavejustice/min-vid/commit/a81deadce48203f2fc847f9a07e39d1f8bb9ba5e

PR #1019, opened 2017-08-28
Add missing deps to get 'npm run package' working

6a68:
I had to install a few things for `npm run package` to finish successfully. Probably you just had them installed globally or something :+1:


meandavejustice:
I don't have any of these installed globally, we aren't using any of these in the package, they are deps of dependencies. Maybe npm messed up initially?

meandavejustice:
Just cloned fresh, ran npm install and npm run package and everything is working without these deps. closing.

Issue #1024, opened 2017-09-01
Can't drag min-vid window


hiseni:
Min-vid window appear to be **non draggable** && **non resizable** on latest Nightly on Linux.
It just sit at the center of my monitor.

This behavior makes so lovely extension completely useless. :cry: 

Please, have a look at this problems.

P.S.: Ping me for some help if needed.
:heart:  this project.

AmrEldib:
Same behavior on Windows 10. Popup appeared behind task bar which I have at the top
I'm using Nightly as well.

SoftVision-PaulOiegas:
@hiseni Hi and thanks for your report. This is a regression caused by Nightly that only reproduces on Linux.

Last good revision: a972e0fb330b7229f86b26dfc75279b44a41c25b
First bad revision: 0ab0630eec0c15b591b379454fe59344bac62290
Pushlog:
https://hg.mozilla.org/integration/mozilla-inbound/pushloghtml?fromchange=a972e0fb330b7229f86b26dfc75279b44a41c25b&tochange=0ab0630eec0c15b591b379454fe59344bac62290

Seems like Bug 1394556 introduced this new behavior.

@meandavejustice, is there anything we can do on our side?
Ciprian also opened Bug 1396512 in Bugzilla for the Foreground issue (#1023) that has the same root cause like this one.

@AmrEldib The issue you have mentioned has other root cause on Windows and it's already tracked under #499.

6a68:
Hi all, thanks for the bug report.

The recent change that landed is not a regression; it enforces strict mode for any scripts loaded using the Firefox JSM importer (`Cu.import`). It seems that a third-party library that we use for platform-specific hacks (@noitidart's ostypes library) violates strict mode, so the loader refused to load the file.

I think #1029 might fix this bug, as well as #1023. (I'm surprised we aren't getting reports of breakage on windows, because ostypes has the same problems on that platform...)

6a68:
Just to confirm, #1029 fixes this bug for me on unity / ubuntu, Dave similarly had good luck with xfce. Should have a fix out shortly

PR #1029, opened 2017-09-05
1023 topify strict mode

jaredhirsch:
@meandavejustice Mind taking a look? This fixes the mac topify behavior for me. (Do you have access to a mac?)

### DIFF
--- a/lib/vendor/noitidart-ostypes/ostypes_mac.jsm
+++ b/lib/vendor/noitidart-ostypes/ostypes_mac.jsm
@@ -14,7 +14,8 @@ if (ctypes.voidptr_t.size == 4 /* 32-bit */) {
 }
 
 var osname;
-if (this.DedicatedWorkerGlobalScope) {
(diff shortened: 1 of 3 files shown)


6a68:
@meandavejustice Added a commit which similarly deals with top-level 'this' global in the ostypes_x11 file. This might fix #1024, could you try it and see?

meandavejustice:
@6a68 Just verified this is working in xfce on ubuntu

meandavejustice:
I'd love to get this merged in the AM, when you get a moment, would you mind squashing these commits and removing the dependency commit? Those deps are child deps and should be installed by their parent dependencies.

6a68:
Sweet! Good timing. I just confirmed that this patch fixes #1024 on Nightly + Ubuntu 16.04 LTS with Unity window manager.

6a68:
@meandavejustice squashed, dep commit removed, good to land

PR #1079, opened 2017-10-08
Fix metrics logging for pause event

vickychijwani:
Hi! I came across a small error while browsing the code, thought a fix might save some confusion for the people looking at analytics data :smile:

EDIT: I guess eslint would have caught this, but `no-unused-expressions` is disabled in [the configuration](https://github.com/meandavejustice/min-vid/blob/master/.eslintrc.json#L44). You may want to consider enabling it, depending on false positive rate.

### DIFF
--- a/frontend/components/player-view.js
+++ b/frontend/components/player-view.js
@@ -93,7 +93,7 @@ export default class Player extends React.Component {
   }
 
   onPause() {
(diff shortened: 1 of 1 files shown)


meandavejustice:
Thanks @vickychijwani ! I've opened https://github.com/meandavejustice/min-vid/issues/1080 to review or eslint rules this week.

PR #1090, opened 2017-10-13
Update icon-overlay's URL

### LINKED ISSUE #1075: Wrong video being played
I click 'Play now' on the second video and I get the first video:
screen shot 2017-10-05 at 4 17 02 pm

Firefox Developer Edition 57.0b5 (64-bit) on macOS 10.12.6

victorhqc:
Youtube's new UI (https://youtube.com/new) updates the DOM elements in a strange way when user click's homepage logo. Instead of deleting the DOM nodes, they are updated in a way that the `min-vid` className is not cleared, so the URL and eventListener are pointing to an obsolete link. This caused a wrong behavior for the new homepage.

Unfortunately for my fix. There will be some seconds when you click the homepage icon, that the URLs will be out of sync. Until the `setTimeout` cicks in (3 seconds at most) and updates the URLs

Fix #1075


meandavejustice:
works great, thanks @victorhqc

victorhqc:
Awesome! :) No problem.

PR #1092, opened 2017-10-19
Shield hackin

6a68:
still barfing on the config passed into studyUtils but the better fix is probably to assign the log key inside the studyUtils object.assign, rather than try to remove it as in this commit


6a68:
don't think this is needed any longer

PR #1095, opened 2017-10-25
unload global imports

meandavejustice:


6a68:
This should fix the problem of modules (particularly `Config.jsm`) getting loaded by version N of the shield study addon, then version N getting uninstalled, version N+1 getting installed, but the stale version of `Config.jsm` from version N being loaded into version N+1 of the addon. This is due to the JSM loader caching modules globally after first load.

PR #1132, opened 2017-11-19
Shield variations

6a68:
Things in this pull request:
- update eligibility check to exclude existing min-vid installs
- randomly pick between three variations (inactive, active, activeAndOnboarding), set the value as a pref which is removed at uninstall
- allow QA to override the variations via a separate pref
  - but note that the variations can't be changed restartlessly. I could add this but didn't see it in the [pioneer enrollment study](https://github.com/mozilla/pioneer-enrollment-study/blob/a8703aca37c04d903782cd4bac2d5fd41616e522/extension/bootstrap.js#L240-L265), so maybe it's not needed
- don't start the webextension if the 'inactive' variation is picked, otherwise start minvid as usual

Still need to do something special for the activeAndOnboarding variation.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -26,9 +26,10 @@ const config = {
   },
   study: {
     studyName: 'min-vid-study', // no spaces, for all the reasons
-    variation: {
-      name: 'non-txp'
-    },
+    // inactive: control group, no minvid
+    // active: regular minvid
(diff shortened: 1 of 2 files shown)


6a68:
Also note that the variation info isn't passed to the webextension in this commit

PR #1151, opened 2017-12-21
restrict content script to youtube, vimeo, and soundcloud

meandavejustice:


ScrPits:
Remove ......

this is the wrong way to hav a Pip....

meandavejustice:
@ScrPits This is temporary because of the current implementation. There is a spec being worked on which will allow pip with a much wider range (we restrict this because it is a test pilot experiment, and it needs to be secure). If ya wanna checkout the spec, you can see it here https://github.com/WICG/picture-in-picture

PR #1159, opened 2018-01-06
Shield onboarding branch

meandavejustice:


meandavejustice:
Just realized we were resetting the seenModal flag to false whenever the browser restarts, about to push a fix for this...

webextension/content-scripts/icon-overlay.js
```diff
@@ -125,9 +125,44 @@ function ytEmbedChecks() {
...
+    browser.runtime.sendMessage({
+      title: 'launch',
+      url: 'https://youtube.com' + url,
+      domain: 'youtube.com',
+      action: getAction(ev)
+    });
+  };
+}
+
+function updateClickIfNeeded(el, { url, props }) {
```
pdehaan: Random question, but is it odd that `registerClick()` expects the `el` and `url` to be passed in as a single object/argument, but `updateClickIfNeeded()` separates those into two arguments?

```js
registerClick({ el, url })
// versus
updateClickIfNeeded(el, { url });
```

webextension/content-scripts/onboarding-modal.js
```diff
@@ -0,0 +1,118 @@
...
+
+  const h2 = document.createElement('h2');
+  h2.textContent = 'We\'ve added a little something to your videos';
+
+  const img = document.createElement('img');
+  img.src = browser.extension.getURL('/onboarding.gif');
+
+  const p = document.createElement('p');
+  p.textContent = `We\'re experimenting with a new feature which allows you to
+  pop out web videos and audio from soundcloud into an always on top view so
```
pdehaan: This should probably be "SoundCloud".

pdehaan: Also, does any of this get localized? I see some \*.properties files above, but didn't see any similar looking translation strings.

webextension/content-scripts/icon-overlay.js
meandavejustice: I agree it is a bit weird, this was from the rebase with master. It's a bit of code from a contributor, we can log a new issue, but probably shouldn't be addressed in this pr https://github.com/meandavejustice/min-vid/blob/master/webextension/content-scripts/icon-overlay.js#L140

webextension/content-scripts/onboarding-modal.js
meandavejustice: That's a good question, it certainly _could_ be localized.

@johngruen @clouserw Do yall know if we should support non english languages for the shield study.

wresuolc: If we have localized content everywhere else, I would.  We have enough time to localize it.  /cc @flodolo

flodolo: As far as I'm aware, we only support English for Shield studies. So, we can translate this string, but I don't expect it to be used, which is suboptimal. Can someone confirm who's targeted by these studies?

wresuolc: @johngruen can speak to that

SoftVision-PaulOiegas:
Shouldn't this be merged on the "shield-branch" so we can test the version most close to current TxP version? 
We are still testing the build from "shield-onboarding-branch". Not sure if this branch is the most up to date since you mentioned in the initial email that you "rebased with master to get most recent changes into the shield study branch".

Or maybe I misunderstood your email and the "shield-onboarding-branch" was the one rebased with the master?

meandavejustice:
@SoftVision-PaulOiegas This is the most up to date this pull request is for it to be merged into `shield-study` branch. Not sure what you are asking beyond that

Issue #1149, opened 2017-12-19
Min Vid player is no longer always on top if the focus is changed while a video/song is loading or playing


SoftVision-CarmenFat:
[Notes]:
- Seems that this is a regression caused by issue [#1228841](https://bugzilla.mozilla.org/show_bug.cgi?id=1228841) (see also [#1145)](https://github.com/meandavejustice/min-vid/issues/1145)

Last good revision: b4d98531124336af6ebd253f9e9fd4aa5a53df30
First bad revision: a90317cd54ba39d47ded48c048bfd695e53c7bcd
Pushlog: https://goo.gl/TL7Czc

[Affected versions]:
- Firefox Nightly 59.0a1 (2017-12-19)
- Min-Vid 0.4.6-dev (built on 12/19/2017)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Latest Min-Vid add-on 0.4.6-dev (built on 12/19/2017) was previously installed on a clean profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites and navigate to any supported platform.
2. Send a video/song to the mini player and choose to play it.
3. Click on the Firefox window.
4. Observe the player.

[Expected result]:
- The player remains on top of the Firefox window.

[Actual result]:
- The player is hidden behind the Firefox window.

[Additional notes]:
- Attached a screen recording of the issue: 

![on background](https://user-images.githubusercontent.com/26000308/34159441-f97b0036-e4d1-11e7-98ce-a58c31ed929d.gif)

imyxh:
Temporary workaround on Ubuntu (and probably some other Linux distributions):
Use dconf-editor or gsettings to edit org.gnome.desktop.wm.keybindings.always-on-top and set it to a key of your choice, like ['<Alt>F5'].
Open Min Vid, and press that key (e.g. <kbd>Alt</kbd>+<kbd>F5</kbd>) while the window is in focus.
I know this is an OS-based solution that doesn't involve Firefox at all, but it might be helpful for other users as a quick fix, or lead to a way to implement always-on-top behavior easily on Linux.

6a68:
I suspect this was fixed by the fix for #1145. @SoftVision-CarmenFat could you try to duplicate?

SoftVision-CarmenFat:
[...]

PR #1197, opened 2018-01-24
Remove obsolete 'conditional catch' statement from ostypes code

### LINKED ISSUE #1145: Can't move window
Popup can't be dragged at all on Firefox 59.0a1 (2017-12-11) (32-bit)

6a68:
Fixes #1145.

Looks weird, but I decided to just stick with tabs, not spaces, as the upstream file does.

Haven't tested this fix myself, but this is the only instance of a conditional catch I've seen in the codebase.


6a68:
also fixes #1149

PR #1243, opened 2018-02-08
clean up modal

johngruen:


meandavejustice:
@johngruen Looks like there is an issue with the positioning of the modal

modal-placement

meandavejustice:
Thanks John!

PR #1281, opened 2018-02-26
make gif smaller

johngruen:


6a68:
@johngruen Good question! I added a commit prior to yours that deletes the gif. Seems to fix the issue. Merging.

PR #1287, opened 2018-03-05
Fix localization placeholders

acid-chicken:
I localized the update of commit 0527765700a5c930739021cab6e3c54a3057c95b.

### Before

https://github.com/meandavejustice/min-vid/blob/f086286a62cc5ce88749651a42a93a6e344f4cfe/locales/en-US/addon.properties#L15

### After

https://github.com/meandavejustice/min-vid/blob/0527765700a5c930739021cab6e3c54a3057c95b/locales/en-US/addon.properties#L26


meandavejustice:
@acid-chicken Our localization is handling with a platform called [pontoon](https://pontoon.mozilla.org/). We don't accept pull requests to localization files directly on the repo. I appreciate the effort! Thank You

PR #1293, opened 2018-03-23
Update placeholders in loading_view_msg

flodolo:
Our French localizer noticed that a string was changed (in #957) without a new ID. As a consequence, we have a ton of locales with the wront placeholders.

Tried to fix Persian and Hebrew via Pontoon, since RTL is a pain in Atom.


flodolo:
@meandavejustice Can you explain why you closed this without any comment?

PR #1320, opened 2018-04-30
Remove detection from non top level frames, sc key and ga key

meandavejustice:


6a68:
we might want to wait for talos data before merging this. the content script changes may not solve the problem, and there's no need to land in github to try things out in mozilla CI.

